PR #13, opened 2019-06-12
Feature/components react button props

eeegor:
Draft implementation of the button interface for `react`.

Points for discussion:
- Is `look` the right name for the functionality (`primary`, `secondary`, `...`)? 
Previously it was named `type` but it collides with `<input type="...">`
- Find the little flaws in the current version of `button` 😉 
- Missing important props?


tomhoefer:
I dont have any comment - shall we already merge it?

PR #15, opened 2019-06-17
Feature/stencil

eeegor:
This initiates the `stencil.js` version of `telements`

TODO:
- Benchmark `lit-html` and `stencil` and choose the right one for the job
- Create separate demos for `stencil` and `lit-html`


examples/boilerplate-react/src/typings.d.ts
```diff
@@ -6,10 +6,10 @@
  */
 
 import * as React from 'react';
-import { IntrinsicTelements } from '@telements/components-web/dist/IntrinsicTelements';
+// import { IntrinsicTelements } from '@telements/components-web/dist/IntrinsicTelements';
```
tomhoefer: Will you uncomment these lines in the future?

packages/components-stencil/readme.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Stencil Component Starter
+
+This is a starter project for building a standalone Web Component using Stencil.
+
+Stencil is also great for building entire apps. For that, use the [stencil-app-starter](https://github.com/ionic-team/stencil-app-starter) instead.
+
+# Stencil
+
+Stencil is a compiler for building fast web apps using Web Components.
```
tomhoefer: Maybe we should add the source of this description, wdyt?

packages/components-stencil/src/components/alert/readme.md
```diff
@@ -0,0 +1,20 @@
+# t-button
```
tomhoefer: I am bit confused why the headline is called t-button but the markdown file is placed within the alert directory

packages/components-stencil/src/components/card/readme.md
```diff
@@ -0,0 +1,21 @@
+# t-button
```
tomhoefer: Same here - shouldnt it be called card instead of t-button?

examples/boilerplate-react/src/typings.d.ts
eeegor: yes, because we currently have 2 web-component setups for exploration
1. Lit-Html (commented out)
2. Stencil

In the PR description `Create separate demos for stencil and lit-html` refers to this point. We need to decide which one to use, and abandon the other...

packages/components-stencil/src/components/alert/readme.md
eeegor: fixed

packages/components-stencil/src/components/card/readme.md
eeegor: fixed

examples/boilerplate-react/src/typings.d.ts
tomhoefer: ok

packages/components-stencil/readme.md
tomhoefer: ok - the file is actually generated by stencil

PR #16, opened 2019-06-18
fix: update jest config path

eeegor:


yarn.lock
```diff
@@ -945,17 +945,10 @@
   resolved "https://registry.yarnpkg.com/@types/jest-diff/-/jest-diff-20.0.1.tgz#35cc15b9c4f30a18ef21852e255fdb02f6d59b89"
   integrity sha512-yALhelO3i0hqZwhjtcr6dYyaLoCHbAMshwtj6cGxTvHZAKXHsYGdff6E8EPw3xLKY0ELUTQ69Q1rQiJENnccMA==
 
-<<<<<<< HEAD
```
tomhoefer: Are these line changes actually intended? and the ones below as well?

PR #38, opened 2019-08-26
Feature/add tag pill and tag links

maomaoZH:


packages/components/src/components.d.ts
```diff
@@ -84,7 +84,11 @@ export namespace Components {
   interface TInputText {}
   interface TTag {
     /**
-    * (optional) Button variant
+    * (optional) Tag on an <a> element
+    */
+    'link'?: boolean;
```
eeegor: link should accept a string f.e.: 'http://some-url.com' or '#some-id-to-jump-to'

packages/components/src/components/tag/tag.tsx
```diff
@@ -8,14 +8,22 @@ import classNames from "classnames";
   shadow: true
 })
 export class Tag {
-  /** (optional) Button variant */
+  /** (optional) Tag variant */
   @Prop() variant?: string = "";
+  /** (optional) Tag on an <a> element */
+  @Prop() link?: boolean = false;
```
eeegor: pill be an own prop - a boolean, and produce a dedicated state class: 'tag--pill'; 
variant prop is for variations like: 'primary', 'secondary', 'danger' etc.

packages/components/src/components/tag/tag.tsx
```diff
@@ -8,14 +8,22 @@ import classNames from "classnames";
...
 
   private getCssClassMap(): CssClassMap {
     return classNames("tag", this.variant && `tag--variant-${this.variant}`);
   }
 
   render() {
+    if (this.link)
```
eeegor: instead of 'this.link' check if the string is not empty (see previous comment, accept a string not a boolean)

packages/components/src/components/tag/tag.tsx
```diff
@@ -8,14 +8,22 @@ import classNames from "classnames";
...
+      return (
+        <a href="#" class={this.getCssClassMap()}>
```
eeegor: ... href="#" ... would return the 'this.link' prop

maomaoZH:
i updated the code based on your comment

packages/components/src/components.d.ts
maomaoZH: hi, i change to string, the code is outdated. can you refresh it?

packages/components/src/components/tag/tag.tsx
maomaoZH: i updated this part this morning. now the condition is to check empty string

packages/components/src/components/tag/tag.tsx
maomaoZH: the code is outdated. can you check updated version

packages/components/src/components/tag/tag.tsx
```diff
@@ -8,14 +8,28 @@ import classNames from "classnames";
...
+  @Prop() pill?: boolean = false;
+  /** (optional) Tag on an <a> element */
+  @Prop() link?: string = "";
 
   private getCssClassMap(): CssClassMap {
-    return classNames("tag", this.variant && `tag--variant-${this.variant}`);
+    return classNames(
+      "tag",
+      this.variant && `tag--variant-${this.variant}`,
+      this.pill && `tag--pill`
```
eeegor: Add 'tag--link' class if link prop is used

packages/components/src/components/tag/tag.css
```diff
@@ -17,3 +25,28 @@
   padding: var(--tag-padding, 0.5rem);
   font-size: var(--tag-font-size, 0.8rem);
 }
+
+.tag--pill {
+  border-radius: var(--tag-pill-border-radius, 0.75rem);
+  padding: var(--tag-pill-padding, 0.4rem);
+}
+
+a.tag {
```
eeegor: instead of targeting 'a.tag' please use a 'tag--link' css class

packages/components/src/components/tag/tag.css
```diff
@@ -17,3 +25,28 @@
   padding: var(--tag-padding, 0.5rem);
   font-size: var(--tag-font-size, 0.8rem);
 }
+
+.tag--pill {
+  border-radius: var(--tag-pill-border-radius, 0.75rem);
```
eeegor: default border-radius should be a little bigger (currently it's not perfectly rounded, run `yarn start` to see preview)

packages/components/src/components/tag/tag.css
```diff
@@ -17,3 +27,33 @@
   padding: var(--tag-padding, 0.5rem);
   font-size: var(--tag-font-size, 0.8rem);
 }
+
+.tag--variant-primary {
+  color: var(--tag-variant-primary-color, #fff);
+  background-color: var(--tag-variant-primary-background, #007bff);
```
eeegor: please use `background` instead of `background-color`

eeegor: Also it makes sense to provide a border variable / default color

packages/components/src/components/tag/tag.css
```diff
@@ -17,3 +27,33 @@
...
+}
+
+.tag--pill {
+  border-radius: var(--tag-pill-border-radius, 1rem);
+  padding: var(--tag-pill-padding, 0.4rem);
```
eeegor: Why do we have smaller padding here then in the default version 🤔 ?

PR #41, opened 2019-08-28
Feature/add tests to badges

steffilankes:


packages/components/src/components/badges/badge.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+		});
+		expect(page.root).toMatchSnapshot();
+    });
+
+
+	it('should have a default css class', () => {
+		expect(element.getCssClassMap()).toBe('badge');
+	})
+
+	it("should have a link", async () => {
```
eeegor: Although you potentially can use a snapshot to test this feature it makes more sense to just test the logic with one unit test:
```
it('should handle link css class', () => {
		element.link = 'http://example-url.com'
		expect(element.getCssClassMap()).toContain('badge--link');
})
```

steffilankes: done

PR #43, opened 2019-09-27
feat/create component toast

SylviaGonschior:
added unit tests


packages/components/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,86 @@
...
+      this.theme && `toast--theme-${this.theme}`,
+      this.variant && `toast--variant-${this.variant}`
+    );
+  }
+
+  private autohideTime = 5000;
+
+  onCloseToast = () => {
+    this.opened = false;
+    console.log("onCloseToast");
```
eeegor: Please remove `console.log` before merge

packages/components/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,86 @@
...
+  @Method()
+  async openToast() {
+    this.opened = true;
+  }
+
+  getTime = () => {
+    const formattedTime =
+      this.time &&
+      formatDistance(subSeconds(this.time, 3), new Date(), { addSuffix: true });
+    console.log(formattedTime);
```
eeegor: Please remove `console.log` before merge

packages/components/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,86 @@
...
+    return formattedTime;
+  };
+
+  onHideToast = () => {
+    if (this.autohide !== false) {
```
eeegor: Currently this event fires even if the `toast` is not opened

packages/components/src/components/toast/toast.tsx
```diff
@@ -0,0 +1,86 @@
...
+      setTimeout(this.onCloseToast, this.autohideTime);
```
eeegor: You need to make sure that the `toast` only can be opened once, and that the timeout is also only fired once. Currently if you call `openToast()` multiple times it creates multiple timeouts which leads to buggy behaviour. Also you'll need to `clearTimeout()` once it's finished or aborted (component not on page anymore e.g. page switch) see https://www.w3schools.com/jsref/met_win_cleartimeout.asp for more details. 

You can debug the timeout doing something similar to this: 
```javascript
if (this.opened && this.autohide !== false) {
    const myTimeout = setTimeout(this.onCloseToast, this.autohideTime);
    console.log('myTimeout', myTimeout)
    return myTimeout;
} else {
    return null;
}
```

test/__mocks__/date-fns.js
```diff
@@ -0,0 +1,3 @@
+export const formatDistance = (date, baseDate) => `${+date} [${baseDate}]`;
+
+export const subSeconds = (date, amount) => `${+date} [${amount}]`;
```
eeegor: `date-fns` should be in `devDependencies`, it's currently missing 😉 

You can add it with `yarn add date-fns --dev`

SylviaGonschior: oh, I did it several times

eeegor: image

still missing...

eeegor: Maybe you forgot to commit ;)

packages/components/src/components/toast/toast.tsx
eeegor: image
the timeout is still firing multiple times..

PR #54, opened 2019-10-15
Removes nextjs yarn link command from build.sh since the next example…

reld:
… boilerplate-next folder does not exists


eeegor:
@reld thanks!

PR #64, opened 2019-11-12
feat: implement component Link

maomaoZH:
https://github.com/telekom/telements/issues/63


packages/components/src/components/link/link.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  public render() {
+    if (!!this.href) {
+      if (!!this.newTag) {
+        return (
+          <a href={this.href} class={this.getCssClassMap()} target="_blank">
+            <slot />
+          </a>
+        );
+      }
```
eeegor: you can omit this condition and add a ternary check below `target={this.newTag ? "_blank" : null}`

packages/components/src/components/link/link.tsx
```diff
@@ -0,0 +1,52 @@
...
+})
+export class Link {
+  /** (optional) link href */
+  @Prop() public href?: string = '';
+  /** (optional) link disabled */
+  @Prop() public disabled?: boolean = false;
+  /** (optional) link variant */
+  @Prop() public underline?: boolean = false;
+  /** (optional) link open a new tag */
+  @Prop() public newTag?: boolean = false;
```
eeegor: the naming `newTag` is a bit confusing...

packages/components/src/components/link/link.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (!!this.href) {
+      if (!!this.newTag) {
+        return (
+          <a href={this.href} class={this.getCssClassMap()} target="_blank">
+            <slot />
+          </a>
+        );
+      }
+      return (
+        <a href={this.href} class={this.getCssClassMap()}>
```
eeegor: you need to check if the link is not disabled, and only then provide the `href` e.g.: `href={!this.disabled && this.href}`, otherwise the disabled link is still clickable

maomaoZH: i set css class -> cursor not allowed.  The link is not clickable

packages/components/src/components/link/link.tsx
maomaoZH: rename to openNewTab

packages/components/src/components/link/link.tsx
eeegor: it actually is :)

PR #93, opened 2020-01-07
feat: add back to top component

maomaoZH:


maomaoZH:
sry @eeegor  still need to update css class name. not ready for review yet.

maomaoZH:
ready to go. thanks

packages/components/src/components/back-to-top/back-to-top.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { CssClassMap } from '../../utils/utils';
+import classNames from 'classnames';
+
+@Component({
+  tag: 't-back-to-top',
+  styleUrl: 'back-to-top.css',
+  shadow: true,
+})
+export class BackToTop {
+  @Prop() public visibilityHeight?: number = 200;
```
eeegor: Please add documentation

packages/components/src/components/back-to-top/back-to-top.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  @State() private visible: boolean = false;
+
+  @Listen('scroll', { target: 'window' })
+  public handleScroll(ev) {
```
eeegor: Please don't use abbreviations e.g. `ev`, use `event` instead

packages/components/src/components/back-to-top/back-to-top.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  public render() {
+    this.visible = document.documentElement.scrollTop > this.visibilityHeight;
+    return (
+      <div
+        class={this.getCssClassMap()}
+        onClick={this.scrollToTop}
+        onScroll={() => this.handleScroll}
+      >
+        <slot />
```
eeegor: The slot is not very usable right now as the `css` is not compatible. If there is a slot, the arrow is still displayed and the scroll button is not adapting the slot content. 
image

maomaoZH:
update code accordingly

PR #106, opened 2020-02-18
refactor: input field

eeegor:
- rename `input text` to `input`
- remove internal validators
- add `input-group` component
- add `input-label` component
- add `input-error` component
- add types `email`, `hidden`, `number` , `password`, `tel`, `text`, `url`


packages/components/src/components/input/input.tsx
```diff
@@ -0,0 +1,63 @@
...
+  @Prop({ mutable: true }) public errorMessage: string;
+  /** (optional) Input text event changed */
+  @Event() public changed: EventEmitter<string>;
+  /** (optional) Input text state touched */
+  @State() public touched: boolean = false;
+
+  public handleChange(event) {
+    this.value = event.target ? event.target.value : this.value;
+    this.changed.emit(this.value);
+    this.touched = true;
```
maomaoZH: hi, may i ask why do we need touched state value?

PR #178, opened 2020-04-14
List component

eeegor:


eeegor:
image

Please make sure that we have the same alignment for list with a custom icon

PR #179, opened 2020-04-16
CircleCI optimized workflows

reld:
This introduces the concept of a release branch so development and master branches stay independent from the deployment process.

We need to create the release branch.

This merge also adds adjustments needed into the root package file and the lerna config file accordingly


package.json
```diff
@@ -18,7 +18,7 @@
         "watch:components": "yarn workspace @scale/components start",
```
reld, line 18: please update the scope name

PR #183, opened 2020-04-21
Symbol instances

kig:
Tracking work on completing symbol instances support.

* Creates an overrideValue for each value that differs between the SymbolMaster and the symbol instance JSON from html2sketch.

### DIFF
--- a/packages/generator-sketch/src/index.js
+++ b/packages/generator-sketch/src/index.js
@@ -2,6 +2,7 @@ const {
   Sketch,
   Page,
   Rect,
+  Style,
   Artboard,
   SharedStyle,
   SymbolMaster,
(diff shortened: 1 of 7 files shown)


kig:
I ran a diff between a broken export (button labels don't show up) and a working one. The working one was created by changing the button symbol master font size to 14 and then back to 16 in Sketch.

The issue seems to be that the group frame of the text is set to width: 0, height: 0. The nested text layer does have the correct width and height.

kig:
The broken frame is coming from html2asketch. The button shadow DOM <slot> returns a zero getBoundingClientRect. Workaround is to use a selection-based BCR if the node BCR is empty in html-to-sketch nodeTreeToSketchGroup.

eeegor:
@kig looking good, thanks!

PR #186, opened 2020-04-23
Pseudo-element support for html-to-sketch

kig:
This PR tracks work to add support for :after, :before and other pseudo elements.

### DIFF
--- a/packages/html-to-sketch/html2asketch/nodeToSketchLayers.ts
+++ b/packages/html-to-sketch/html2asketch/nodeToSketchLayers.ts
@@ -54,12 +54,116 @@ function isSVGDescendant(node: HTMLElement) {
   return (node instanceof SVGElement) && node.matches('svg *');
 }
 
+const PSEUDO_ELEMENTS = [':after', ':before'];
+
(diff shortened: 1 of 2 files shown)


kig:
What it does: When nodeToSketchLayers comes across a :before or :after pseudo element (getComputedStyle(node, ':before')), it creates a new span and copies the pseudo element computed style to the span, then adds the span to the start or end of the node's children. The content of the :before / :after rule is added as the textContent of the new span (attr, url, etc. are not supported at this point.)

::-webkit-slider-thumb and ::-webkit-slider-running-track require some extra work, as they exist in the INPUT element's shadow DOM. The INPUT element shadow DOM is closed, i.e. the shadowRoot is null. So you can't descend into the INPUT element shadow DOM.

To work around this, nodeToSketchLayers tests if the node matches to any of the document's style rules that have the ::-webkit-slider-thumb / ::-webkit-slider-running-track selector. If the node has a slider style rule, nodeToSketchLayer creates a DIV, initializes it with Chrome's default slider styles, and copies the found style rule on top of it. The DIV is then added to the parent of the INPUT node (INPUT's children are not visible, need to add as sibling) and positioned on top of the INPUT node. To make add-node-to-parent work, the nodeTreeToSketchGroup function needs to check if new children were added after processing a node's children, and process any found new children.

PR #199, opened 2020-04-28
feat: create first iteration of possible sidebar navigation component

acstll:
As discussed these past days, here's a “prototype” of sidebar navigation.

> 🙋‍♂️ For simplicity and to save time I’m temporarily using things like regular CSS and custom properties that do not belong in the current strategy

I followed this example, just a bit, again mainly for a11y:
https://inclusive-components.design/collapsible-sections/ (It works nicely with the keyboard.)

The whole thing is composed of 3 components:

- `sidebar-nav` (the main container)
- `sidebar-nav-item` (basically a `<li>`) 
- `sidebar-nav-collapsible` (a `<li>` that accepts a list of `sidebar-nav-item`s and does the toggling)

The reason for `sidebar-nav-item` to be only a `<li>` is to allow using any kind of link components provided by frameworks (Next.js, Gatsby, etc.).

## Usage example

```html
<scale-sidebar-nav>
  <!-- single link -->
  <scale-sidebar-nav-item>
    <a href="/a">Schlitz sustainable</a>
  </scale-sidebar-nav-item>

  <!-- collapsible block -->
	<scale-sidebar-nav-collapsible label="Heading with h2">
    <scale-sidebar-nav-item>
      <a href="/b">Literally fashion axe</a>
    </scale-sidebar-nav-item>
    <scale-sidebar-nav-item>
      <a href="/c">Four loko health</a>
    </scale-sidebar-nav-item>
  </scale-sidebar-nav-collapsible>
</scale-sidebar-nav>
```

## Screenshot

It looks like this with a bit of external styling.

Captura de Pantalla 2020-04-28 a les 16 42 48

## Meta

There's this other thought I'd like to share with you: maybe this—"sidebar navigation"—is too big for a component and it could be more like a pattern you should be able to build or compose out of other more "primitive" components the design system provides. Are we sure this is within the scope of a design system as we're currently thinking of it?

## Request for comments

Chatting with Tim Heiler I got the following feedback, which I’m going to list below:

- is using headings (h2, h3, etc.) inside navigation the right thing to do?
- would it be possible to make this behave like a stack-sliding menu “for mobile” without changing the markup, only by using some kind of context flag?
- we should be able to use `<scale-icon>` to customize the arrow

I’d have some more question regarding some details, but I’d like to get your feedback first.

With a bit of refactoring and cleaning up it should be usable in Home OS.

What do you think? is this the way to go?


packages/components/src/index.html
```diff
@@ -610,5 +610,80 @@ <h3>Breadcrumb - custom icon separator</h3>
...
+        
+        <scale-sidebar-nav-collapsible label="Heading with h2">
+          <scale-sidebar-nav-item>
+            <a href="/b">Literally fashion axe</a>
+          </scale-sidebar-nav-item>
+          <scale-sidebar-nav-item>
+            <a href="/c">Four loko health</a>
+          </scale-sidebar-nav-item>
+
+          <scale-sidebar-nav-collapsible label="Nested heading with h3" level="3">
```
eeegor, line 660: Can we automate the `level`? Maybe be a bit error prone to manually have to keep track of it

packages/components/src/components/sidebar-nav/sidebar-nav-collapsible.styles.ts
```diff
@@ -0,0 +1,33 @@
+import { JssStyle } from 'jss';
+
+export const styles: JssStyle = {
+  collapsible: {
+    margin: 0,
+  },
+  heading: {
```
eeegor, line 7: So far we have used `BEM` style for classes - we need to discuss if we want to continue using this approach. One benefit: With `BEM` we can export static `CSS` stylesheets which could be used without `web-components`

packages/components/src/components/sidebar-nav/sidebar-nav-collapsible.tsx
```diff
@@ -0,0 +1,104 @@
...
+  shadow: true,
+})
+export class SidebarNavCollapsible implements Base {
+  /** (optional) Injected jss styles */
+  @Prop() styles?: StyleSheet;
+  /** decorator Jss stylesheet */
+  @CssInJs('SidebarNavCollapsible', styles) stylesheet: StyleSheet;
+
+  @Prop() label: string;
+  @Prop() level: number = 2; // TODO validate this is a num between 2 and 6
```
eeegor, line 20: Let's have a thought if we can automate `level`

packages/components/src/components/sidebar-nav/sidebar-nav-collapsible.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const expanded = button.getAttribute('aria-expanded') === 'true';
+    const target = button.parentElement.nextElementSibling;
+    button.setAttribute('aria-expanded', !expanded);
+    target.hidden = expanded;
+  }
+
+  render() {
+    // This is basically a reset
+    const hostStyles = `
+      button {
```
eeegor, line 39: You could put this into the `JSS` sheet: 

```javascript
collapsible: {
  ...
  '& button': { ... },
  '& h2, & h3, & h4, & h5': { ... }
}
```

packages/components/src/components/sidebar-nav/sidebar-nav-collapsible.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const { classes } = this.stylesheet;
+
+    return (
+      <Host>
+        <style>
+          {hostStyles}
+          {this.stylesheet.toString()}
+        </style>
+        <li class={this.getCssClassMap()}>
+          <HeadingTag class={classes.heading}>
```
eeegor, line 71: Should we allow for manual override e.g. prop: `heading`?

packages/components/src/components/sidebar-nav/sidebar-nav-collapsible.tsx
```diff
@@ -0,0 +1,104 @@
...
+            <button
+              class={classes.button}
+              onClick={this.handleClick}
+              aria-expanded="false"
+            >
+              {this.label}
+              <svg
```
eeegor, line 78: Please use the `scale-icon` like you suggested in your comment 😉

packages/components/src/components/sidebar-nav/sidebar-nav-item.styles.ts
```diff
@@ -0,0 +1,9 @@
+import { JssStyle } from 'jss';
+
+export const styles: JssStyle = {
+  item: {
```
eeegor, line 4: `BEM` ... see comment above

packages/components/src/components/sidebar-nav/sidebar-nav.styles.ts
```diff
@@ -0,0 +1,10 @@
+import { JssStyle } from 'jss';
+
+export const styles: JssStyle = {
+  nav: {},
```
eeegor, line 4: `BEM` see comment above

PR #205, opened 2020-05-03
feat/table

nowseemee:


packages/components/src/components/table/table.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  componentWillUpdate() {}
+
+  render() {
+    // on initial render
+    if (!this.slots.table) {
+      // build object of slots
+      // @ts-ignore - fromEntries should be fine here
+      this.slots = Object.fromEntries(
+        Array.from(this.hostElement.children).map(v => [v.slot, v])
```
eeegor, line 36: Please avoid shortcuts 😉: instead of `...map(v => [v.slot, v])` better do `...map(child => [child.slot, child])`

packages/components/src/index.html
```diff
@@ -610,5 +610,106 @@ <h3>Breadcrumb - custom icon separator</h3>
       <a href="/b">Breadcrumb 2</a>
       <span>Breadcrumb 3</span>
     </scale-breadcrumb>
+
+    <h3>Table</h3>
+
+    <script>
+      function sort(element) {
```
eeegor, line 617: Can we somehow make this user friendlier? Maybe an internal method will do?

packages/components/src/index.html
```diff
@@ -610,5 +610,106 @@ <h3>Breadcrumb - custom icon separator</h3>
...
+    <scale-table variant="regular">
+      <table slot="table">
+        <thead>
+        <div slot="header" style="padding: 20px 8px 0px 8px;">
+          <scale-text variant="h5" slot="header">Table title</scale-text>
+        </div>
+        <tr>
+          <th aria-sort="descending" onclick="sort(this)">Title</th>
+          <th aria-disabled="true" onclick="sort(this)">Tags</th>
+          <th onclick="sort(this)">Stats</th>
```
eeegor, line 633: `sort` is currently not actually sorting...

PR #222, opened 2020-05-12
chore/table-follow-up

nowseemee:
Done:
- add docs
- remove complex slots
- styles
- rename variant to size

[how it should look like](https://user-images.githubusercontent.com/4532334/81727454-48bd0900-9489-11ea-9476-b44db4ecd586.png)


packages/components/src/components/table/table.tsx
```diff
@@ -16,55 +16,37 @@ export class Table implements Base {
   @Prop() customClass?: string = '';
   /** (optional) Display sort arrows on/off */
   @Prop() showSort?: boolean = false;
-  /** (optional) Visual variant */
-  @Prop() variant?: 'regular' | 'compressed' = 'regular';
+  /** (optional) Visual size */
+  @Prop() size?: 'small' | 'big' = 'small';
```
eeegor, line 20: Please use `large` instead of `big`

nowseemee: I'm going to use `small` and `default`.

nowseemee: OK!

PR #228, opened 2020-05-14
Update README.md

maomaoZH:


README.md
```diff
@@ -18,7 +18,7 @@ This repository uses `lerna` and `typescript`
 cd path-to/folder-name
 git clone git@github.com:telekom/scale.git
 cd scale
-git checkout development
+git checkout master
```
eeegor, line 21: Actualy we can remove this line I think, as `master` will be checked out by default 🙂

maomaoZH: ok. i will remove it.

maomaoZH: removed it

PR #232, opened 2020-05-18
Feat/textare

maomaoZH:


packages/components/src/components/input/input.styles.ts
```diff
@@ -102,6 +111,11 @@ export const styles: JssStyle = {
       },
     },
   },
+  'input--disable-resize': {
+    '& .input__textarea': {
+      resize: 'none',
```
eeegor, line 116: We need to expand this to allow for multiple settings, as resize could be: `unset | none | vertical | horizontal`

Maybe we could achieve it through the component and using the `jss` `stylesheet.addRules`?

packages/components/src/components/input/input.tsx
```diff
@@ -58,6 +63,8 @@ export class Input implements Base {
   @Prop() counter?: boolean;
   /** (optional) radio checked value */
   @Prop() preChecked?: boolean;
+  /** (optional) textarea disableResize */
+  @Prop() disableResize?: boolean;
```
eeegor, line 67: Check comment above :)

eeegor:
The vertical position of the text is not correct when used with an animated label...

image

eeegor:
When hovering the `textarea` there is no transition on the border, please check the `input` as a reference 🙂

maomaoZH:
The transition effect is removed when i removed transition property because when we resize the textarea, it does not look so good

Issue #237, opened 2020-05-29
@scaleds/docs


reld:
After publishing the package and using/installing it from npm the package throws an error.
I think webpack is not handling JSX right for some reason.
I will look further into this.

@nowseemee maybe you can check this too?

cheers

PR #239, opened 2020-06-05
fix/monument

nowseemee:
fix: decouple from @scaleds/components
fix: transpile @scaleds/docs
feat: dynamic whitelist aka featuredComponents config property

### DIFF
--- a/packages/docs/gatsby-config.js
+++ b/packages/docs/gatsby-config.js
@@ -6,6 +6,12 @@ module.exports = {
   },
   plugins: [
     {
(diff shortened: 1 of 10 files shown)


packages/docs/gatsby-config.js
```diff
@@ -5,6 +5,12 @@ module.exports = {
     author: `@telekom`,
   },
   plugins: [
+    {
+      resolve: `gatsby-plugin-compile-es6-packages`,
+      options: {
+        modules: [`@scaleds/docs`]
```
nowseemee, line 11: this fixes #237

PR #270, opened 2021-04-26
build: publish to npm

nowseemee:


package.json
```diff
@@ -14,15 +14,19 @@
     "bootstrap": "lerna bootstrap",
     "postbootstrap": "lerna run generate",
     "new-version": "lerna version --conventional-commits --no-push",
-    "force-version": "lerna version --conventional-commits --force-publish --no-push",
-    "format": "lerna run format"
+    "force-version": "lerna version --conventional-commits --force-publish",
+    "format": "lerna run format",
+    "publish-telekom": "lerna publish -y from-package --no-git-tag-version --no-push --force-publish @telekom/scale-components,@telekom/scale-components-react,@telekom/scale-components-angular,@telekom/scale-components-vue,@telekom/scale-design-tokens",
+    "prepare-neutral": "node scripts/prepare-neutral-release.js && git commit . -m 'chore: temp commit'",
```
acstll, line 20: what happens to this commit? 🤔 does it get lost/piled up in the CI machine?

acstll (commented):
everything else looks 🆗 to me

package.json
nowseemee: it goes to the void 🤷🏻

nowseemee: would you like to keep it? if yes, where would you store it?

acstll: nope, the opposite, if it goes into the void then it's fine 😄

PR #283, opened 2021-04-28
fix: tokens for headings

acstll:
- [x] update token values for headings


acstll:
@nowseemee what's the normal thing to do to update the visual tests? asking for a friend… 🤓

nowseemee:
> @nowseemee what's the normal thing to do to update the visual tests? asking for a friend… 🤓

Your friend needs to have docker running on their computer - just install Docker Desktop from docker.com, then run : `yarn workspace @telekom/scale-visual-tests test -u` and profit 🎉

acstll:
I'm updating the visual tests in another PR (I'm unable to do so right now), and the "diff" is actually tiny — the headings in the data grid 👀

PR #284, opened 2021-04-28
fix: add story for readonly attribute in text-field, also attribute/story for textarea

acstll:
- [x] "readonly" story for text-field
- [x] "readonly" attribute and story for textarea (🍒 bonus)

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -34,7 +34,7 @@
     "format": "prettier --write \"*.ts\" \"src/**/*.{html,tsx,ts,css,json}\" \"!src/**/*.d.ts\"",
     "lint:fix": "yarn lint:style --fix",
     "lint": "tslint -c tslint.json -p . \"src/**/*.{tsx,ts,css,json}\" \"!src/**/*.d.ts\"",
-    "develop": "stencil build --docs --watch -c stencil-dev.config.ts",
+    "develop": "stencil build --docs --watch",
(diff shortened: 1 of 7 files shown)


packages/components/package.json
```diff
@@ -34,7 +34,7 @@
     "format": "prettier --write \"*.ts\" \"src/**/*.{html,tsx,ts,css,json}\" \"!src/**/*.d.ts\"",
     "lint:fix": "yarn lint:style --fix",
     "lint": "tslint -c tslint.json -p . \"src/**/*.{tsx,ts,css,json}\" \"!src/**/*.d.ts\"",
-    "develop": "stencil build --docs --watch",
+    "develop": "stencil build --docs --watch -c stencil-dev.config.ts",
```
nowseemee, line 37: what's the difference between "dev" and "develop" scripts? can we have just one?

acstll: awesome question, I would also want to have just one as well

the difference is the `--serve` flag that triggers the dev server:
- `dev` is used for developing using html pages, quick and dirty
- `develop` is used for running along side Storybook in watch mode, so you can develop/test in Storybook, slow and dirty haha

I'm gonna remove this from here, figure it out so we can have one that serves both purposes, and open another PR, ok?

nowseemee: ok :)

PR #315, opened 2021-04-30
fix: broken `min` and `max` attributes in Slider (and more)

acstll:
## The problem

Setting the `max` attribute to anything higher than `100` makes the inner bar break outside of its container.

Also, the value displayed might not be needed (it's not gonna be always *percentages*, and then anything other than `100` makes no sense…), or there should be an option to ~~hide it or even~~ modify the "unit" (%).

Screenshot of a native `input type=range` and the `scale-slider` going off, both with a `max` value of 360:
slider going off with a `max` value of 360

## Expected behaviour

More like a native `<input type=range>`.

## In this PR

- [x] simplified inner logic (less code, removed a few `@State` props that were triggering extra renders)
- [x] `min` and `max` attributes works as expected
- [x] added `unit` attribute to change the default "%" or leave it out
- [x] added `decimals` attribute to control how many decimals to show (0, 1 or 2)
- [ ] ~~renamed `show-value` to `hide-value` to invert the logic, so default is `false` and it behaves more like a classic boolean attribute 🔴~~
- [x] added the `scaleInput` event, that fires on every input change, and changed `scaleChange` event to fire only when user input stops (native-like)
- [x] updated tests and snapshots

### ~~🔴 Breaking~~

~~The only breaking change here would be `show-value` being changed to `hide-value`. If we undo this, it should be safe as non-breaking.~~

### Example application in `src/html/slider.html`

Run `yarn start` to see in action.

image


nowseemee (approved):
<3

Issue #316, opened 2021-05-02
Accordion Header not updating


tommy-codez:
Hi there,
it seems there is a bug or strange behavior in the accordion component. (Didn't check other components yet)

I use react and a translation functionality with (react-i18next). With normal behavior the change of language causes a rerender of all the text where the translation hook is invoked. (t-function)

here the beginning of an example code

```
import React from "react";
import { useTranslation } from "react-i18next";

import {
  ScaleAccordion,
  ScaleCollapsible,
} from "@telekom/scale-components-react";

const MyAccordion = (props) => {
  const { t } = useTranslation();

  return (
    <div>
      {t("sm1")}
      <ScaleAccordion>
        <ScaleCollapsible>
          <h1>{t("sm1")}</h1>
          <p>
            {t("sm1")} Freegan kinfolk farm-to-table humblebrag cred, hammock
            bespoke small batch pabst…
          </p>
```
after the language changes every instance of t("sm1") gets correctly translated (the one before the accordion and the one inside the p-tag - except the one in the header (h1) .. any idea why - or is there a workaround?

acstll:
Thanks again @tommy-codez for reporting this.

This is a very interesting one. The accordion component uses a "progressive enhancement" approach ([based on this](https://inclusive-components.design/collapsible-sections/)) that does some direct DOM manipulation in the light DOM, if I remember correctly; that should be causing this unexpected behaviour.

Let's see how we can fix this one. (A possible workaround until we have a fix would be wrapping the accordion in an extra React component, to make sure it completely re-renders on each language change?)

tommy-codez:
Hi there again :)
when the fix is official I will test it again if it works with the translation library. As I said we will use the scale components in our project in the upcoming time and I will report bugs here if we find some :)
Greetings

PR #319, opened 2021-05-03
fix: remove direct DOM manipulation in Collapsible

acstll:
This is meant to solve #316 

We explicitly have to set the heading text using a slot, so updates in the light DOM work as they're suppose to… Maybe not as nice but it works without quirks.

```diff
- <h2>Leo integer malesuada nunc vel risus</h2>
+ <span slot="heading">Leo integer malesuada nunc vel risus</span>
```

Before:

```html
<scale-collapsible>
  <h2>Leo integer malesuada nunc vel risus</h2>
  <p>Shoreditch raclette knausgaard, swag enamel pin food truck everyday carry 3 wolf moon.</p>
</scale-collapsible>
```

After:

```html
<scale-collapsible>
  <span slot="heading">Leo integer malesuada nunc vel risus</span>
  <p>Shoreditch raclette knausgaard, swag enamel pin food truck everyday carry 3 wolf moon.</p>
</scale-collapsible>
```

~~(Marking it WIP because I want to double check this is the right thing.)~~

---

_Update:_ previous syntax still works so this is no breaking change 👐


nowseemee (approved):
🙏🏻

PR #323, opened 2021-05-05
fix(select): add  -moz query for rendering visual padding correctly

oddcelot:
fixes additional spacing for selected input value generated by firefox


oddcelot:
oh, can someone check if the rendering issue may be OS related?
win/linux version of FF might render just fine 🤷‍♂️

PR #326, opened 2021-05-07
Sketch/checkbox symbols

oddcelot:


nowseemee:
do i see changes overlapping with `feat/sketch-icons`? I'll review this once we are through that then, ok?

Issue #328, opened 2021-05-10
Switch and Table not providing Styles Attribute


Lalaluka:
Hey,
i noticed that Scale Switch and Table do not provide styles Attributes like every other element.
Is there a specific reason for that? Will that change in the future?

It's a bit weird to have it a different way for just two components.

nowseemee:
The styles prop is to be removed since we have css variables and the shadow parts.

Lalaluka:
So instead of adding styles to _table_ and _switch_ it would help more to remove Shadow DOM from other components?

nowseemee:
We should remove the `styles` prop from all components - shadow DOM should stay as it is ✌️

nowseemee:
Fixed in https://github.com/telekom/scale/pull/378

PR #339, opened 2021-05-16
fix: remove unused dependencies move used ones to dev

Lalaluka:
Hey,
I noticed that the vue-components are delivered with jest and vue Test libraries as dependencies, which are not needed in the delivered files.
So I removed:

- vue-class-component
- vue-template-compiler

completely since they are never used.

And moved:

- vue-jest
- @vue/test-utils

to devDependencies since they are only used during testing.

This leaves just 'vue' as a needed dependency, which may be changed to a peerDependency in future Versions?

Tests and Compile do pass. 
To test the functionality of the Package I also included my branched components into a default vue2 Project with the current scale-components from npmjs and tested the components Button and Switch. Which still works like intended.

If these tests are not sufficient let me know and I can do some more work on this pr.

### DIFF
--- a/packages/components-vue/package.json
+++ b/packages/components-vue/package.json
@@ -26,17 +26,15 @@
   ],
   "devDependencies": {
     "@testing-library/jest-dom": "^5.8.0",
+    "@vue/test-utils": "^1.2.0",
     "jest": "^26.1.0",
     "rimraf": "^2.6.3",
     "ts-jest": "^26.2.0",
-    "typescript": "^3.3.4000"
+    "typescript": "^3.3.4000",
+    "vue-jest": "^3.0.7"
   },
   "dependencies": {
-    "@vue/test-utils": "^1.0.3",
-    "vue": "^2.6.11",
-    "vue-class-component": "^7.2.3",
-    "vue-jest": "^3.0.5",
-    "vue-template-compiler": "^2.6.11"
+    "vue": "^2.6.11"
   },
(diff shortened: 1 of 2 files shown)


nowseemee:
Hi Lalaluka, excellent - I'll make sure this is included in today's release. Many thanks for this improvement. I prepared another PR where the diff on yarn.lock is lighter: https://github.com/telekom/scale/pull/339

nowseemee:
It's released under `3.0.0-beta.9` - thanks a lot!

Lalaluka:
Hey,
Thanks for (kinda) accepting my pr.
For future PRs I guess yarnpkg is your preferred registry?
Just wondering, because Scale itself is mainly published to npmjs and just mirrored to yarnpkg. But I can keep it in mind in the future.

nowseemee:
Hi Lalaluka, thanks for the question - yes, yarnpkg is the preferred registry. I figure it was rather rude to close your PR in favor of mine instead of asking you to change it. Sorry for the bad experience and I hope I'll have the chance to actually accept a PR of yours in the future. Thanks again for your contribution!

Lalaluka:
Don't worry. A clean dependency Tree is more rewarding than an accepted PR.
Currently working on #328 so another PR might happen ;)

nowseemee:
Yeah awesome! Actually we want to remove the styles prop from all the components so if you feel like, you could really help us out here.

PR #344, opened 2021-05-18
refactor: remove styles prop from all components

### LINKED ISSUE #328: Switch and Table not providing Styles Attribute

oddcelot:
Closes #328


nowseemee:
closing in favor of https://github.com/telekom/scale/pull/378

PR #345, opened 2021-05-18
fix(components): incorrect esm module link in README

acstll:
The vanilla snippets currently use a broken URL, this fixes that.

Captura de Pantalla 2021-05-18 a les 13 12 31


nowseemee:
Nice catch! Can you scan the whole repo for the same mistake, please?

acstll:
yep 🤓

acstll:
much better now, thank you @nowseemee !

PR #346, opened 2021-05-18
Feat/checkbox group

marvinLaubenstein:


packages/components/src/components/checkbox-group/checkbox-group.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { newSpecPage } from '@stencil/core/testing';
```
ChrisPaj, line 1: maybe it is possible to raise the coverage up to 60%?

PR #370, opened 2021-05-28
refactor(components): use margin instead of padding in tab-panel

acstll:
This allows "margin collapse", so the vertical spacing does not visually break when _unstyled_ block elements with some by-default margins, like `p`, `ul`, etc. are used inside `scale-tab-panel`.

(Some experts suggest avoiding "margin collapsing" completely, because it's hard to manage. But I think in this case it's pretty much fine.)


acstll:
I don't think visual-tests need to be updated for this one, the `diff` only shows something in checkbox 🤓

PR #372, opened 2021-05-28
Update includes and symbol names

oddcelot:


oddcelot:
@nowseemee any idea why the `accordion` suddenly fails the visual test? (seems to be the overall height 🤔)

PR #379, opened 2021-06-01
chore: adjust documentation showStatus no default

Lalaluka:
Hi,
while working with the progress bar I noticed another problem with the documentation. It suggests that show-status has a default value, which is not the case in the Stencil Component. Also, the shown Code for the first and second examples doesn't match the canvas.

I changed the default Value of show-status to none in the Storybook-Vue file so it does not suggest that show-status has a default.
Also, I changed the first example to be coherent to before adding `show-status: true` to the canvas and adding it to the example HTML code.
The second example now does not show the status anymore, but the example HTML code also doesn't contain it.

Like this, the Storybook represents the Stencil component and the example codes match the canvases they represent.

The Storybook is building and is working correctly. No additional tests.

If not the documentation is wrong, but the Stencil Component has unwanted behavior, I can change it the other way around in another PR, I already prepared such a Branch.

### DIFF
--- a/packages/storybook-vue/stories/3_components/progress-bar/ProgressBar.stories.mdx
+++ b/packages/storybook-vue/stories/3_components/progress-bar/ProgressBar.stories.mdx
@@ -150,7 +150,8 @@ For Shadow Parts, please inspect the element's #shadow.
     args={{
       percentage: 20,
       customColor: "#009DE0",
-      label:"Progress custom"
+      label:"Progress custom",
+      showStatus: true
     }}
   >
     {Template.bind({})}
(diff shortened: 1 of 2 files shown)


Lalaluka:
I noticed that the visual test broke because my commit affected the second canvas by removing the default (which was intended).

So I restored the second example to `show-status` and the first canvas now does not show the status text.
Like this, the Visual Test passes again. 

I think this is the better way of doing it instead of changing the visual-tests. Also the functionality of `show-status` is tested like this.

nowseemee:
Looks good, thanks a lot!

Issue #382, opened 2021-06-02
Icons not visible at http://localhost:3333


ChrisPaj:
When starting the scale project preview I can not see any icons on index.html at http://localhost:3333/

acstll:
@ChrisPaj can you post a screenshot? I see some icons, but others are not showing because of out-dated syntax.

We should definitely clean up this index.html file.

Lalaluka:
I guess you mean the Buttons?

![Bildschirmfoto 2021-06-04 um 15 34 56](https://user-images.githubusercontent.com/26364692/120809543-766c6880-c54a-11eb-9b9d-391ae5564470.png)

Index-html Line 585
```html
<h3>Button icon Before</h3>
<scale-button
  icon-before="M10 2a8 8 0 016.472 12.703l4.912 4.913.06.065a1.25 1.25 0 01-1.726 1.794l-.102-.091-4.913-4.912A8 8 0 1110 2zm0 1.5A6.508 6.508 0 003.5 10c0 3.584 2.916 6.5 6.5 6.5s6.5-2.916 6.5-6.5-2.916-6.5-6.5-6.5z">
  Hello
</scale-button>

<h3>Button icon only</h3>
<scale-button
  icon="M10 2a8 8 0 016.472 12.703l4.912 4.913.06.065a1.25 1.25 0 01-1.726 1.794l-.102-.091-4.913-4.912A8 8 0 1110 2zm0 1.5A6.508 6.508 0 003.5 10c0 3.584 2.916 6.5 6.5 6.5s6.5-2.916 6.5-6.5-2.916-6.5-6.5-6.5z">
</scale-button>
```
The Syntax is outed and the button doesn't use the icon Prop anymore but should use Icons in the `slot` e.g.
```
<scale-button icon-only="true">
  <scale-icon-navigation-external-link accessibility-title="External link, opens in new tab"></scale-icon-navigation-external-link>
</scale-button>
```

I can write a quick PR for the Button Icons. I could not find any other occurrences of missing icons personaly.

acstll:
thank you @Lalaluka 🙏

PR #389, opened 2021-06-04
fix: icon Buttons on Indexhtml wrong syntax

Lalaluka:
Tries to fix the problem I found with #382 

### Before
Bildschirmfoto 2021-06-04 um 15 34 56

### Now
Bildschirmfoto 2021-06-04 um 15 58 23

Uses the same example code as the Storybook.

### DIFF
--- a/packages/components/src/html/index.html
+++ b/packages/components/src/html/index.html
@@ -573,12 +573,10 @@
     <h3>Button variants</h3>
     <scale-button
       variant="primary"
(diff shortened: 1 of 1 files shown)


nowseemee (approved):
<3

nowseemee (changes requested):
oops, prettier failed - can you reformat please, @Lalaluka?

Lalaluka:
Hi @nowseemee ,
sry for the missing format, I should have noticed that error.
But it's fixxed now.

To mitigate errors like that in the future, would it be interesting to extend the husky with a precommit function?

nowseemee:
Hi @Lalaluka, thanks for making the build pass! I don't want to force all devs to have those hooks running before every commit or push of theirs. Here is an opt-in way to set it up on your computer: https://prettier.io/docs/en/precommit.html#option-5-shell-script

PR #390, opened 2021-06-07
fix: remove child nodes, which are whitespace strings

ChrisPaj:
child nodes, which are whitespace strings have to be removed to let 'icon after' work fine

### DIFF
--- a/packages/components/src/components/button/button.tsx
+++ b/packages/components/src/components/button/button.tsx
@@ -85,13 +85,10 @@ export class Button {
    * If so, it's probably an icon, so we set `iconPosition` to `after`.
    */
   setIconPositionProp() {
-    const nodes = Array.from(this.hostElement.childNodes);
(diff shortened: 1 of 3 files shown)


packages/components/src/components/button/button.tsx
```diff
@@ -86,6 +86,12 @@ export class Button {
    */
   setIconPositionProp() {
     const nodes = Array.from(this.hostElement.childNodes);
+    // delete empty text nodes, which are due to formatting
```
acstll, line 89: I would change "delete" to "skip" or "ignore", because "delete" might give one the impression that it's altering the DOM, when it's only about the array 😅

acstll (changes requested):
Thank you @ChrisPaj, this is great!

Do you think it would also be possible to add a unit test for this case?

acstll (approved):

PR #400, opened 2021-06-11
Feat/rating stars fixes

ChrisPaj:


packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -24,30 +39,37 @@ export class RatingStars {
...
+  @Event() scaleChange!: EventEmitter<void>;
 
   colorFilled = `var(--scl-color-primary)`;
   colorBlank = `var(--scl-color-grey-50)`;
   size = this.small ? '16px' : '24px';
 
-  getSymbol = (color: string, size: string, selected?: boolean) => {
+  @Watch('value')
+  handleValueChange() {
+    this.scaleChange.emit();
```
acstll, line 69: It'd be handy if we send the value in the event here: `emit({ value: this.value })`

acstll, line 69: This is a very clean an elegant technique (watching for changes in `value`), but it has a drawback: it can create a feedback loop (we made that mistake in the past haha).

`set value -> trigger event -> listen to event and update value -> set value -> ...`

So it's a bit more verbose emitting events in many places, but events should only be emitted after direct user interaction (mouse, touch).

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -141,37 +191,52 @@ export class RatingStars {
...
     return Math.ceil(numberToRound * multiplier) / multiplier;
   }
 
   getAriaLabel() {
-    return this.ariaTranslation
-      .replace(/\$\{x\}/gi, this.rating.toString())
-      .replace(/\$\{y\}/gi, this.numOfStars.toString());
+    return this.ariaLabelTranslation
+      .replace(/\$\{x\}/gi, this.value.toString())
+      .replace(/\$\{y\}/gi, this.max.toString());
```
acstll, line 220: I would use `value` and `max` here intead of `x` and `x`, like: `${value} out of ${max} stars`, since it's more explicit.

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -141,37 +191,52 @@ export class RatingStars {
         this.precision
       ),
       0,
-      this.numOfStars
+      this.max
     );
     return star;
   }
 
+  getValueFromXCoordinate(coordinate: number) {
```
acstll, line 199: This is just a suggestion because it doesn't make the component better, only the code :)

Maybe we can also use this function to handle the mouse input? The code is more or less the same as in `getValueFromMousePosition`.

acstll (changes requested):
Thank you very much @ChrisPaj, this is great!

Apart from the notes:
- you can also update the `index.html` file to replace the `rating` attribute.
- 1 test is failing on my machine (some unhandled promise rejection), does that happen to you as well?

Thanks again :)

ChrisPaj:
> Thank you very much @ChrisPaj, this is great!
> 
> Apart from the notes:
> 
> * you can also update the `index.html` file to replace the `rating` attribute.
> * 1 test is failing on my machine (some unhandled promise rejection), does that happen to you as well?
> 
> Thanks again :)

Thank you very much for the review @acstll :-)
I have changed  the `index.html` file to replace the `rating` attribute by `value`
On my machine no test is failing

packages/components/src/components/rating-stars/rating-stars.tsx
ChrisPaj: The emitter now emits the actual value

ChrisPaj: I have dropped the watcher. The emitter is now called on user interaction

packages/components/src/components/rating-stars/rating-stars.tsx
ChrisPaj: ...done

packages/components/src/components/rating-stars/rating-stars.tsx
acstll: I really like the explicitness of the function name `emitRatingValue()` :))

packages/components/src/components/rating-stars/rating-stars.tsx
ChrisPaj: Mouse and Touch are now using `getValueFromXPosition(evTou?: TouchEvent, evKey?: MouseEvent)` This function could easily have one parameter only (ev: any), but strange enough some unit test is in need of the explicit type (Mouseevent). It is an alternative to rewrite the function with one parameter only and to drop this very unit test

acstll: I think the function is fine with the two params!

`evTou` looks a bit funny 😄, again explicitness is preferred so I would have used `evTouch` or even `touchEvent`. But we can leave this like that.

acstll (approved):

Issue #405, opened 2021-06-15
Difference between Scale Icon and Scale IconLib


Lalaluka:
Hi,
I just noticed a difference between the `<scale-icon>` Element and the Scale Icon library.
Scale-icon uses the `fill` Attribute to color an icon while the Library has the `color` Attribute.

Code Examples:
```
<scale-icon-alert-security color="green"></scale-icon-alert-security>
<scale-icon fill="green" path="M12...75h8.7z"></scale-icon>
```

Is there a specific functional reason for the difference between `fill` and `color`?

If yes it might be helpful to state it in the Storybook somewhere. Or at least make the Icon Library Tab interactive (at the moment it's static).

If not I think it would be nice to have one Attribute name for the functionality e.g. add color to `scale-icon` or fill to the IconLibrary.

Lalaluka:
I think it's not worth opening another Ticket for it but for the icon _device-device-tv_ the **selected** svg from B&D has been uploaded instead of the default one.
Default:
<img width="65" alt="Bildschirmfoto 2021-06-15 um 12 50 06" src="https://user-images.githubusercontent.com/26364692/122040543-37ef6d00-cdd8-11eb-843e-9ae24cc2db4f.png">

Scale: 
<img width="65" alt="Bildschirmfoto 2021-06-15 um 12 51 07" src="https://user-images.githubusercontent.com/26364692/122040657-5bb2b300-cdd8-11eb-81c9-907f4e0d9c08.png">

I can't fix it with a quick PR, because I think the icon Download from Brand&Design is bugged?
[B&D Download](https://www.brand-design.telekom.com/asset/icon-for-user-interfaces-0-icons-fr-user-interfaces/)
I can't retrieve the full IconLibrary just partial Icons 🤔
Maybe someone else has access to a full zip :)

acstll:
hey @Lalaluka thanks for this!

> Is there a specific functional reason for the difference between fill and color?

the reason is that they do kinda different things, `scale-icon` is mostly a wrapper, you can use any `path` with it so being able to set `fill` or `stroke` independently made sense. This isn't true the Telekom's icon library, you don't need such granularity, so `color` seemed enough.

I can understand it's not consistent and that's not nice.

Renaming an attribute would be a breaking change though, and we would like to avoid that at the moment.

[...]

PR #412, opened 2021-06-16
feat: add fill to icon with higher priority than color

Lalaluka:
Hi ✌️
yesterday I opened #405 

Resulting from that I took a look at what the icon handlebar actually does (icon-component.hbs). And as you can see in line 44 of the icon component, it just adds the `color` Prop as an SVG `fill` attribute.

This is the same behavior the scale-icon Wrapper Component has (over the pathAttributes).

So I added a `fill` prop which has priority over the `color` prop. Resulting in this behavior:
```
<scale-icon-action-circle-close color="green"> </scale-icon-action-circle-close>
<scale-icon-action-close fill="red"> </scale-icon-action-close>
<scale-icon-action-download fill="red" color="green"> </scale-icon-action-download>
```
Bildschirmfoto 2021-06-16 um 23 04 06

I reorded the Storybook a bit so it meets the behavior. Showing examples with `fill` instead of `color`.
Like this scale-icons and icons from the Library have the same behavior using `fill`, without removing the support for `color`.

Tests and Visual Tests are passing for me (out of accordion failing occasionally but that also happens on main).

### DIFF
--- a/packages/components/scripts/icon-component.hbs
+++ b/packages/components/scripts/icon-component.hbs
@@ -20,9 +20,9 @@ export class {{className}} {
   /** (optional) The width and height in pixels */
   @Prop() size?: number = 24;
   /** (optional) Sets the icon color via the `fill` attribute */
-  @Prop() color?: string = 'currentColor';
-  /** (optional) Sets the icon color via the `fill` attribute */
   @Prop() fill?: string = 'currentColor'; 
+  /** (optional) Alias for `fill` */
+  @Prop() color?: string = 'currentColor';
   /** (optional) If `true`, the icon changes visually */
   @Prop() selected?: boolean = false;
   /** (optional) If `true` the SVG element will get `aria-hidden="true"` */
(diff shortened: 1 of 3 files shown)


packages/components/scripts/icon-component.hbs
```diff
@@ -20,7 +20,9 @@ export class {{className}} {
   /** (optional) The width and height in pixels */
   @Prop() size?: number = 24;
   /** (optional) Sets the icon color via the `fill` attribute */
```
acstll, line 22: Since both `color` and `fill` do the same thing, I'm thinking it might end up being confusing for maintainers, and even consumers.

I would try and document `color` strictly as an alias for `fill`. Something like:

```js
/** (optional) Sets the icon color via the `fill` attribute */
@Prop() fill?: string = 'currentColor'; 
/** (optional) Alias for `fill` */
@Prop() color?: string = 'currentColor';
```

acstll (changes requested):
Thank you @Lalaluka !

There is one little suggestion, let me know what you think.

packages/components/scripts/icon-component.hbs
Lalaluka: That looks great!

That's a good addition

acstll (approved):
👌

PR #429, opened 2021-06-25
chore: add better doc for variant and size to Tags

Lalaluka:
Hey,
I noticed in the Storybook that the Chip/Tags don't have a selector for `variant` and `size` like for example Buttons.

For Buttons it looks like this:
Bildschirmfoto 2021-06-25 um 16 39 55
vs
Bildschirmfoto 2021-06-25 um 16 40 31
for Tags

Also, I noticed that size and variant is per default not specified and has no enumerating type like Buttons have.
Button.tsx Line 25:
`@Prop() size?: 'small' | 'large' = 'large';`
vs Tag.tsx Line 21:
`@Prop() size?: string = '';`

So I changed that and also added a select field to the Storybook. Now it looks like this:
Bildschirmfoto 2021-06-25 um 17 00 45

I'm thinking of defining the `default` variant/size (or primary) in the component too. It would not be a breaking change but would improve maintainability in my view.
Let me know if this should be added to this PR.

### DIFF
--- a/packages/components/src/components/tag/tag.tsx
+++ b/packages/components/src/components/tag/tag.tsx
@@ -100,7 +100,7 @@ export class Tag {
   getCssOrBasePartMap(mode: 'basePart' | 'css') {
     const component = 'tag';
     const prefix = mode === 'basePart' ? '' : `${component}--`;
-    
+
     return classNames(
       mode === 'basePart' ? 'base' : component,
       this.size && `${prefix}size-${this.size}`,
--- a/packages/components/src/components/tag/tag.tsx
+++ b/packages/components/src/components/tag/tag.tsx
@@ -18,9 +18,9 @@ import classNames from 'classnames';
 })
 export class Tag {
   /** (optional) Tag size */
-  @Prop() size?: '' | 'small' = '';
+  @Prop() size?: 'small';
   /** (optional) Tag variant */
-  @Prop() variant?: '' | 'secondary' = '';
+  @Prop() variant?: 'secondary';
   /** (optional) Tag href */
   @Prop() href?: string = '';
   /** (optional) Tag target */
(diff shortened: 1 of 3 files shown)


packages/components/src/components/tag/tag.tsx
```diff
@@ -18,9 +18,9 @@ import classNames from 'classnames';
 })
 export class Tag {
   /** (optional) Tag size */
-  @Prop() size?: string = '';
+  @Prop() size?: '' | 'small' = '';
```
nowseemee, line 21: do we really need to define an empty string here as a possible value? the `?` makes the prop optional already. I think it would make sense to change this to `@Prop() size?: 'small';` What do you think?

packages/components/src/components/tag/tag.tsx
```diff
@@ -18,9 +18,9 @@ import classNames from 'classnames';
...
   /** (optional) Tag variant */
-  @Prop() variant?: string = '';
+  @Prop() variant?: '' | 'secondary' = '';
```
nowseemee, line 23: I think the same here - `?` makes the prop optional so I don't see why we need an empty string.

nowseemee (commented):
Hi Calvin @Lalaluka, many thanks for opening this PR. I posted two questions of the same nature - would you be so kind to take a look?

packages/components/src/components/tag/tag.tsx
Lalaluka: Actually no we won't need this. That's a good addition.
Just checked if the classNames function would work correctly with a possible `undefined` state of size but it does.

packages/components/src/components/tag/tag.tsx
Lalaluka: Same here works fine like this.

Lalaluka:
Could have been one commit instead of three but here we go :D

nowseemee (approved):

nowseemee:
Great, thanks! 👍🏻

PR #432, opened 2021-07-01
docs(storybook): separate header from shell

nowseemee:


acstll:
the "separate header from shell" headline is so nice! :)

PR #434, opened 2021-07-01
chore: implement Node password-protected staging server, tidy up scripts

acstll:
There's a new Node static server inside scripts/, one should be able to run it from the root folder, and previously had built Storybook, also possible from the root.

To try it out:

- run `yarn build:storybook:staging`
- run `HTTP_USERNAME=admin HTTP_PASSWORD=1234 yarn serve:storybook:staging`

Whatcha think @nowseemee?

---
Update: I would have preferred not adding any more new deps in the root, but making a new package for this seemed less preferable.


reld:
I like this 👍

nowseemee:
Lovely 👍🏻

Issue #440, opened 2021-07-08
Overlapping Label/Value in DateField Component


tommy-codez:
Hi there,

I am not sure if this a react specific issue or if its reproducable.

If the value of date Field is set with a state value
 ( for example value={state.validFrom} ) 
the label/value is overlapping.

![grafik](https://user-images.githubusercontent.com/66920189/124890007-e5206400-dfd7-11eb-9045-52abaf17f35a.png)

This doesnt happen when the value is set with a direct value - like 
(value={"2021-07-07"})

I assume there is a problem with rerendering ?!

greetings,

Lalaluka:
I can't reproduce your behavior😅
I tried a function component and a class component with:
React 17
Scale/React-components both: 3.0.0-beta.15

My Class-based component looks like this:
```
import './App.css';
import React, { Component } from 'react';
import { ScaleDatePicker } from '@telekom/scale-components-react';


class App extends Component {
  constructor(props) {
    super(props)
    this.state = {
      test: "2020-12-31"
    }
  }
  render() {
    return (
      <div className="App">
        <scale-date-picker label="Helper Text" value={this.state.test}></scale-date-picker>
        <ScaleDatePicker label="Helper Text" value={this.state.test}></ScaleDatePicker>
      </div>
    );
  }
}

export default App;
```

Both simple WebComponents and the React extensions work as intended.
![Bildschirmfoto 2021-07-08 um 11 54 20](https://user-images.githubusercontent.com/26364692/124902682-747f4480-dfe3-11eb-9649-9bc7081f49af.png)

Do you have the Scale CSS set properly? Your Fonts don't look like TeleNeo to me.
Or do you have some example code?

tommy-codez:
Thank you @Lalaluka  for your support. I will recheck the fonts/CSS, but it should be TeleNeo.

We use the date fields in a modal dialog where the initial value are empty strings, but the state changes during opening - perhaps its a timing issue ?! - I will recheck

```
 const openNewModal = () => {
    var tzoffset = emptyCompanyCode.validFrom.getTimezoneOffset() * 60000;
    var localISOTime = new Date(
      emptyCompanyCode.validFrom.getTime() - tzoffset
    );

    setState({
      ...state,
      modalOpen: true,
      companyCode: emptyCompanyCode,
      errors: emptyValidationErrors,
      validFrom: localISOTime.toISOString().split("T")[0],
      validTo: "",
      created: "",
      modified: "",
[...]

PR #441, opened 2021-07-08
fix: datepicker not updating on outside value change

Lalaluka:
Hi 🧑‍💻

This is a simple fix for: #440 

Added a watcher to the Value for outside changes of `value`. 
Since the `hasValue` State doesn't seem to have reactivity on its own.

I'm not absolutely keen if StencilJS has some way to make States reactive directly, if there is a possibility to handle it directly in the State that might be nicer. But I couldn't find such functionality.

Can be tested like this:
```
<scale-date-picker label="Helper Text" id="Test" value=""></scale-date-picker>
<script>
  setTimeout(()=>{
    document.getElementById('Test').setAttribute("value", "2020-12-31")
  }, 6000)
</script>
```

This also enables Stuff like this:
```
setTimeout(()=>{
  document.getElementById('Test').setAttribute("value", "2020-12-31")
}, 6000)
setTimeout(()=>{
  document.getElementById('Test').setAttribute("value", "")
}, 8000)
setTimeout(()=>{
  document.getElementById('Test').setAttribute("value", "2020-12-31")
}, 10000)
```

Let me know what you think :)

### DIFF
--- a/packages/components/src/components/date-picker/date-picker.tsx
+++ b/packages/components/src/components/date-picker/date-picker.tsx
@@ -18,6 +18,7 @@ import {
   Event,
   EventEmitter,
   State,
+  Watch,
 } from '@stencil/core';
 import { DuetDatePicker as DuetDatePickerCustomElement } from '@duetds/date-picker/custom-element';
 
@@ -181,6 +182,14 @@ export class DatePicker {
    */
   @Method() async hide(moveFocusToButton = true) {
     return this.duetInput.hide(moveFocusToButton);
+  }
+
+  /**
+   * Watch Value property for changes from outside and change hasValue based on that
+   */
+  @Watch('value')
+  onValueChange() {
+    this.hasValue = this.value != null && this.value !== '';
   }
 
   componentWillLoad() {
--- a/packages/components/src/components/date-picker/date-picker.tsx
+++ b/packages/components/src/components/date-picker/date-picker.tsx
@@ -185,7 +185,7 @@ export class DatePicker {
   }
 
   /**
-   * Watch Value property for changes from outside and change hasValue based on that
+   * Watch `value` property for changes and update `hasValue` based on that.
    */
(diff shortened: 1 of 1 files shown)


packages/components/src/components/date-picker/date-picker.tsx
```diff
@@ -183,6 +184,14 @@ export class DatePicker {
     return this.duetInput.hide(moveFocusToButton);
   }
 
+  /**
+   * Watch Value property for changes from outside and change hasValue based on that
```
acstll, line 188: I would change the comment a bit, for consistency with the rest of the file, something like this:

```
Watch `value` property for changes and update `hasValue` based on that.
```

acstll (changes requested):
This is amazing @Lalaluka, thank you very much for this fix!

I left a picky comment, if you're so kind.

Thanks again!

acstll:
By the way, `@State()` in Stencil is "reactive": it will trigger a render whenever it changes.

From the [docs](https://stenciljs.com/docs/state):

> Any changes to a @State() property will cause the components render function to be called again.

Is that what you had doubts about?

packages/components/src/components/date-picker/date-picker.tsx
Lalaluka: Agreed makes it more readable too

Lalaluka:
Hey @acstll ,
yea `State` is reactive in the way that the Component is rerendered if State is changing.

What I meant was going in the direction of reference. But the longer I think about the less reasonable it is.
You know how in js Objects are references?
```
let a = {a: 1}
let b = a
a = {a: 2}
console.log(b) // {a: 2}
```
Something like that, but for the state value set.
What I didn't think about was that variables are not and state is a variable :)

Don't mind the comment, It is not that smart as I initially thought 😅

acstll (approved):
Thank you once more!

PR #442, opened 2021-07-09
Refactor/rating stars

ChrisPaj:
complete refactoring of rating stars to make them accessible to screen readers


packages/components/src/components/rating-stars/rating-stars.css
```diff
@@ -10,81 +10,151 @@
...
-  position: absolute;
-  bottom: var(--scl-spacing-32, 2rem);
+  display: inline-flex;
+  flex-direction: column;
 }
 
-.rating--disabled .rating__label {
-  color: var(--scl-color-grey-50, #7c7c7c);
+[part='wrapper']:hover .icon-clip {
+  --decimal-width: 100% !important;
```
acstll, line 26: is this `--decimal-width` actually doing anything? since the `precision` prop is gone but maybe there's something I'm missing… 🤔

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -9,298 +9,195 @@
...
+  ratingStarId = `scale-rating-star-${ratingStarCount++}`;
+
+  /** The lower limit of the rating. In cases where  */
+  @Prop({ reflect: true }) starSize: 'small' | 'large' = 'large';
+  /** The lower limit of the rating. In cases where  */
+  @Prop({ reflect: true }) minRating = 0;
+  /** The upper limit of the rating */
+  @Prop({ reflect: true }) maxRating = 5;
+  /** Represents the current value of the rating */
+  @Prop({ mutable: true, reflect: true }) rating = 0;
```
acstll, line 52: I wonder what happened to this prop being called `value`. The reason for having `value` is consistency with the rest of the "form control" components (and form controls in general, it's like a convention) and also playing nice with frameworks (like Angular) that expect the property holding the value to be called `value` in their binding mechanisms.

I'm not asking now that you change it back, I just would like to know the reasoning behind the change :)

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -9,298 +9,195 @@
...
-    statusNote({ source: this.hostElement, tag: 'beta' });
-    this.handleMouseEnter = this.handleMouseEnter.bind(this);
-    this.handleMouseLeave = this.handleMouseLeave.bind(this);
-    this.handleMouseMove = this.handleMouseMove.bind(this);
-    this.handleMouseClick = this.handleMouseClick.bind(this);
-    this.handleKeyDown = this.handleKeyDown.bind(this);
-    this.handleTouchStart = this.handleTouchStart.bind(this);
-    this.handleTouchEnd = this.handleTouchEnd.bind(this);
-    this.handleTouchMove = this.handleTouchMove.bind(this);
+    this.firstStarSelected = false;
```
acstll, line 68: if not intended, this might trigger an unnecessary render. I would set the default value while declaring it above, no?

```ts
@State() firstStarSelected: boolean = false;
```

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -9,298 +9,195 @@
...
   }
 
-  componentDidLoad() {
-    handleListeners(this.element, 'addListeners');
+  // constructs the aria message for the current rating
+  getRatingText() {
+    const filledText = this.ariaLabelTranslation
+      .replace(/\$rating/g, `${this.rating}`)
+      .replace(/\$maxRating/g, `${this.maxRating}`);
```
acstll, line 75: I like this `$` approach; it's less confusing than using the template literal syntax without actually using template literals like before…

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -9,298 +9,195 @@
...
-      this.value = clamp(ratingMinus, 0, this.max);
-      this.emitRatingValue();
-      event.preventDefault();
+  handleTouchEnd = (ev: TouchEvent) => {
+    const star = ev.composedPath()[0] as StarInterface;
+    const starValue = Number(star.dataset.value);
+    if (starValue === 1) {
+      if (this.firstStarSelected) {
+        this.rating = starValue;
+        this.firstStarSelected = !this.firstStarSelected;
```
acstll, line 98: are you sure `firstStarSelected` needs to be `State` and not a simple private property?

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -9,298 +9,195 @@
...
+            id={`${this.ratingStarId}-label`}
+            part="label"
+            htmlFor={this.ratingStarId}
+          >
+            {this.label}
+          </label>
+        )}
+        <input
+          disabled={this.disabled || this.readonly}
+          part="range-slider"
```
acstll, line 176: keyboard support is flaky, when you reach the component by tabbing, keyboard control works (on MS Edge), but after I click, then it doesn't. I'm thinking that maybe the focus target is different in each case… could that be?

acstll (changes requested):
The idea of using an `input type=range` is lovely. I like the approach, it's simple.

Also, I really like this idea of using `[part=""]` as selector for the styles. It's less code (no classes) and also makes you think about parts, so you don't forget… I wonder if there are any hidden drawbacks?

Were you able to fulfil all the accessibility requests in the end? 

Great job 💪

packages/components/src/components/rating-stars/rating-stars.css
```diff
@@ -10,81 +10,151 @@
...
 :host {
-  --rating-spacing: 0.125rem;
-}
+  --scl-stars-color: var(--scl-color-primary);
+  --scl-stars-hover-color: var(--scl-color-primary-hover);
+  --scl-stars-active-color: var(--scl-color-primary-active);
+  --scl-stars-placeholder-color: var(--scl-color-black);
+  --scl-stars-placeholder-revert-color: var(--scl-color-black);
+  --scl-stars-disabled-color: var(--scl-color-grey-50);
+  --scl-stars-spacing: var(--scl-spacing-2);
```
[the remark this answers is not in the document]
ChrisPaj: ok, done

packages/components/src/components/rating-stars/rating-stars.css
```diff
@@ -10,81 +10,151 @@
...
 
-.rating__label {
-  position: absolute;
-  bottom: var(--scl-spacing-32, 2rem);
+  display: inline-flex;
+  flex-direction: column;
```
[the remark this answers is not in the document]
ChrisPaj: I've built a container for the rating stars...

packages/components/src/components/rating-stars/rating-stars.css
ChrisPaj: you are right, this one was no longer needed ;-)

packages/components/src/components/rating-stars/rating-stars.tsx
ChrisPaj: I have changed firstStarSelected to be a private property with a default value of false

packages/components/src/components/rating-stars/rating-stars.tsx
ChrisPaj: I have changed firstStarSelected to be a private property with a default value of false

packages/components/src/components/rating-stars/rating-stars.tsx
ChrisPaj: I have changed this behaviour in a way that on clicking a star the input gets the focus. Now, when clicking a star the arrow keys are still functional.

packages/components/src/components/rating-stars/rating-stars.tsx
acstll: We've discussed this with @oddcelot, and we don't agree haha but it's clear 👍

packages/components/src/components/rating-stars/rating-stars.tsx
acstll: I wasn't able to check this but sounds great, if it comes back (hopefully not), we'll notice 🤓

PR #446, opened 2021-07-15
Feat/pagination accessibility

ChrisPaj:
adaptions due to brand & design requirements


packages/components/src/components/pagination/pagination.tsx
```diff
@@ -87,7 +87,8 @@ export class Pagination {
   @Watch('totalElements')
   calculateWidth() {
     // calculate max possible width
-    this.maxWidth = (this.totalElements.toString().length * 3 + 3) * 9;
+    this.maxWidth = (this.totalElements.toString().length * 4 + 3) * 9;
```
acstll, line 90: I know this is not yours, but these "magic" numbers are a bit intriguing, do you know whey they are like that?

packages/components/src/components/pagination/pagination.tsx
```diff
@@ -87,7 +87,8 @@ export class Pagination {
...
+    console.log(this.maxWidth)
```
acstll: you forgot to remove this one :)

packages/components/src/components/pagination/pagination.tsx
acstll: I was wondering if something like `white-space: nowrap;` in CSS would help?

acstll (changes requested):
Small thing, one question 🤓

packages/components/src/components/pagination/pagination.tsx
ChrisPaj: I don't know how it came to this formula, but it seems to work brilliantly :-)
It is a good idea to give the info part, the one with the numbers a fixed width, because every button click could change this. I went back to the original formula and used white-space: nowrap; Thank you for the hint!

packages/components/src/components/pagination/pagination.tsx
ChrisPaj: removed... :)

PR #463, opened 2021-07-19
Tooltip Documentation

Lalaluka:
Hi @marvinLaubenstein ,
this is a draft of the Tooltip documentation.
I will have to clean up my fork a bit since it's a bit outdated and GitHub shows more code changes than I actually did.

I will add English tomorrow. But some feedback on the german one is appreciated.
I'm not happy with the way the Pictures look. Will probably replace that one too.


Lalaluka:
Replaced with #465

PR #473, opened 2021-07-23
refactor: use http-server for Storybook staging

acstll:
Storybook's build does not work well with some well-known static servers. It fails to load `iframe.html` with both [`serve`](https://github.com/vercel/serve) and Express' [`serve-static`](https://github.com/expressjs/serve-static).

Fortunately, the one Storybook likes, [`http-server`](https://github.com/http-party/http-server), offers basic HTTP auth. So that's what we're using here.

I'm keeping the recent `staging-server.js` script for now. But we can delete that for sure.


nowseemee:
Great! Many thanks for this!

acstll:
I cherry-picked in berners…, will do the same in neptune.

PR #474, opened 2021-07-26
feat: add `list` option to ScaleTextField

gaurishkamat:


gaurishkamat:
@acstll could you please review this one?
Thanks:)

acstll:
Yes, I will be glad to, tomorrow or the day after.

@gaurishkamat thanks a lot!

gaurishkamat:
Hi @acstll any update on this?

acstll:
@gaurishkamat not yet, I'm very sorry for the delay. I will take a look next week.

Do you think it'd be possible to leave the `datalist` element outside of the component?

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -72,6 +72,8 @@ export class TextField {
   @Prop() inputId?: string;
   /** (optional) input background transparent */
   @Prop() transparent?: boolean;
+  /** (optional) input list */
+  @Prop() list?: string | Array<Record<'value' | 'displayValue', string>>;
```
acstll, line 76: I wouldn't take data in the `list` attribute, but keep it simple as take the id of the `<datalist>` element with should be linked with.

That means we can skip the `<datalist>` element and all the `readData` implementation.

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -142,6 +152,7 @@ export class TextField {
       this.status === 'error' ? { 'aria-invalid': true } : {};
     const helperTextId = `helper-message-${i}`;
     const ariaDescribedByAttr = { 'aria-describedBy': helperTextId };
+    const datalist = { 'list': 'data-list' };
```
acstll, line 155: I would put the conditional check here and not below when we spread the variable. Something like this:

```tsx
const list = this.list ? { 'list': this.list } : null;

return <input {...list} />
```

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -172,6 +184,14 @@ export class TextField {
             {...(this.helperText ? ariaDescribedByAttr : {})}
           />
 
+          { this.list && (<datalist id="data-list">
```
acstll, line 187: I would remove this and let the user define it themselves. Mainly to keep things simple, and also because ids need to be unique (in valid HTML), so there cannot be more than one "data-list" id in the document.

So people using the library would do it like this:

```html
<scale-text-field label="List of things" list="things"></scale-text-field>
<datalist id="things">
  <!-- options here -->
</datalist>
```

acstll (changes requested):
Hey @gaurishkamat 

thanks for this and thank you for including unit tests! And sorry once more for the delay.

For the sake of simplicity I'm asking for some changes. Please let me know if anything is unclear.

Thanks again.

packages/components/src/components/text-field/text-field.tsx
gaurishkamat: Hi @acstll,
Thanks for your suggestions. I did not get enough time to look into this. I will work on the review changes in a while.

Thanks again!

PR #482, opened 2021-07-29
Feat/pagination accessibility

ChrisPaj:
approved version concerning accessibility requirements


acstll (approved):
Comments for next time 🤓 

- we don't need the `index.html` changes since we have it [empty already in `main`](https://github.com/telekom/scale/blob/main/packages/components/src/html/index.html).
- I don't quite like the magic numbers, now with decimals (3.5), but we can live with that…
- I also would have liked to use [a spacing token](https://github.com/telekom/scale/blob/main/packages/design-tokens/src/telekom/tokens.js#L34-L35) instead of the `10px` value.

Issue #467, opened 2021-07-21
User Feedback Tabnavigation


Lalaluka:
Hi,
in the landline diagnosis of Deutsche Telekom we worked with Tagnavigation and decided to not use the Scale Implementation (for now). But we would like to give you some feedback on what the issues for us were. And we would be happy to maybe implement one or two new features around the issues.

## 1. Linewrapping
If you have a lot of entries in the Tab Navigation it wraps into a new line like this:
<img width="575" alt="Bildschirmfoto 2021-07-21 um 10 44 41" src="https://user-images.githubusercontent.com/26364692/126459567-f632528e-0a37-4565-9b05-967c97dde053.png">
While that might look nice for some applications it's not usable for us.

Also, I've gotten reports about some weird glitchy behavior when selecting and unselecting tabs in Linewrapped Tab Navigations. But I can't provide any reproducible code.

We would prefer the possibility to disable Linewrapping and use Navigation buttons instead (including a soft scroll on click).
Our own implementation looks like this now:
![tabnav](https://user-images.githubusercontent.com/26364692/126460554-8c0c7692-7eb9-4306-abc6-1d07fc3dd25e.gif)

With Navigation IconOnly Buttons on the Left and Right.

## 2. Disabling Tabs
The first issue wasn't that big of a dealbreaker. But the current Tab Navigations doesn't provide a possibility to diable Tabs. Such functionality is pretty useful for use in Diagnosis Software. 
I think such a functionality would be pretty easy to add too. You can see our approach in the gif too, with the "Reservierung der IP Adresse" being disabled.

acstll:
Thank you @Lalaluka for taking the time to provide this valuable feedback.

I definitely see these as features that could be added to the library. We will discuss them with the rest of the team.

Regarding disabling tabs, you basically `disable` the button inside the header (and make sure the panel is also unreachable by keyboard), or is there something else to it?

Thanks again!

Lalaluka:
Just let me know whats fine to add :)
I will gladly help to implement it.

Exactly we are adding `disabled` to it and it's unreachable through clicks and keyboard Navigation.

Lalaluka:
Hey @acstll ,
do you have any updates on this topic? ✌️

acstll:
Hey, not yet 😬 (we're a little low on resources at the moment ☀️)

Thanks for asking. I will get back to you next week. 🙏

timheiler-ico:
Hi you 2!

[...]

PR #484, opened 2021-07-29
feat: disabled tabs

Lalaluka:
Related to #467 
Implements Tabs that are disableable through the `disabled` attribute.
Usage examples are added to the Storybook and Visual Tests.

I also took a look at the keyboard navigation and it should work as intended (Arrow Keys, END, HOME).


acstll:
Also taking a look at this… 👀

packages/components/src/components/tab-header/tab-header.tsx
```diff
@@ -65,9 +69,9 @@ export class TabHeader {
     return (
       <Host
         id={`scale-tab-header-${this.generatedId}`}
-        role="tab"
+        role={this.disabled ? false : 'tab'}
         aria-selected={String(this.selected)}
```
acstll, line 73: On disabled tabs we get an `aria-selected="undefined"`. I think changing this line to `aria-selected={this.selected ? 'true' : 'false'}` will fix this. What do you think?

acstll (changes requested):
There is one small thing I found (left a comment), other than that this is perfect.

Thanks so much. (Sorry that it took me so long to test it and do the actual review.)

packages/components/src/components/tab-header/tab-header.tsx
Lalaluka: Hi,
yup I agree that `aria-selected="undefined" ` is not that nice. I'll change it in a sec :)

(no worries about the delay ^^ It's vacation time)

Lalaluka: Ok, this breaks the unit tests. I need to take a closer look

acstll: Probably because the snapshots already had "undefined" in them?

```diff
- <scale-tab-header aria-selected="undefined" id="scale-tab-header-0" role="tab" tabindex="-1">
+ <scale-tab-header aria-selected="false" id="scale-tab-header-0" role="tab" tabindex="-1">
```

Maybe `yarn test --spec --updateSnapshot` solves it…

acstll: Looking forward to what you find out. Thanks for the update :)

Lalaluka: Jup that's it. I will update the Snapshot

acstll (approved):

Issue #483, opened 2021-07-29
Option to colorize every second rows in tables


Kaawo:
Greetings everyone,

it would be great if you could a colorize of every second row in the component "table" (like in your beta Data Grid). We received feedback from various users on this. :)
If needed @Lalaluka can implement it.

Here is an example for a design: 
![04_53_53](https://user-images.githubusercontent.com/88150166/127514748-1cbecf34-1dd7-40ca-a0d7-59ca4ee5135b.jpg)

Thanks!

acstll:
Hello @Kaawo 

This sounds great, please go ahead 👍 

Thank you 🙏

acstll:
Implemented in #489 🎉 

Thanks a lot, once more.

We should be publishing a new beta release soon.

PR #489, opened 2021-07-30
feat: striped tables

Lalaluka:
Hi 🧑‍💻,
Here is the Implementation of #483 
I added the prop `striped` to the tables adapting the naming of Bootstrap.
It colors every second Row with grey 0. And can be changed through the css-variable:
`--background-secondary-tbody`
I'm not that happy with the Variable name (just changed it again) would --background-striped better?

Also added to Visual Tests, Storybook and Sketch Library :)

Another thing the Screenshot from @kaawo also had vertical borders. Those are also not present in the Table implementation atm


Lalaluka:
I also noticed some empty rulesets in some of the CSS files.
Are they intended to stay or should they be deleted?

acstll:
Thank you @Lalaluka!

I'm checking this out and testing it today, looks perfect at first look.

For the variable name, here is the convention I drafted some time ago but didn't make it in the docs:

* We use the following naming convention: `--{property/thing}-{description/variant}-{state}`, e.g. `--color-primary-hover`.
* The separation between property/thing, description/variant and state should be evident when reading the variable name.
* For the `property/thing`, we may try the token's category/section first (e.g. `spacing` instead of `margin`), if it doesn't work or it makes more sense, we use the actual property.
* Use `x` for left/right, `y` for top/bottom properties.

Based on this, I would name it `--background-row-striped` or `--background-tr-striped` since there are some variables using `th`?

packages/storybook-vue/stories/3_components/table/Table.stories.mdx
```diff
@@ -247,6 +248,7 @@ scale-table {
   --box-shadow-th-sortable-focus: inset 0 0 0 var(--scl-spacing-2) var(
       --scl-color-focus
     );
+  --background-secondary-tbody: var(--scl-color-grey-0);
```
acstll, line 251: I would probably name this `--background-tr-striped` :)

packages/storybook-vue/stories/3_components/table/Table.stories.mdx
```diff
@@ -380,3 +382,77 @@ Please switch to the Canvas view for a working example
   }
 </script>
 ```
+
+## With striped rows
+
+<Canvas withSource="none">
+  <Story
+    name="With striped rows"
```
acstll, line 390: Let's have the title capitalised, for consistency: "With Striped Rows" 🙏

acstll (commented):
I left some comments, the variable name and something minor.

Thank you very much for updating the hbs template and the visual tests!

packages/storybook-vue/stories/3_components/table/Table.stories.mdx
Lalaluka: Adapted in last commit :)

packages/storybook-vue/stories/3_components/table/Table.stories.mdx
Lalaluka: Thanks for the hint! adapted

acstll (approved):
🚀

acstll (approved):
👍

Kaawo:
Good work. I would like to thank both of you. :)

PR #507, opened 2021-08-05
Feat/toggle group improvements

ChrisPaj:
recommended changes: fix singleSelect story, add story with icons, make selected icon solid, renamings and css stuff


acstll (approved):
From a code point of view, I think this is good to go. Let's wait to QA/design 🤓

timheiler-ico:
Its about the stories:

– can the default border be "on" please?

– i want to have only stories for either the button or the group. Technically, the <scale-toggle-button> is only a sub component and must not be used outside the group.

– story naming is not aligned with properties any more (story "secondary" shows button-variant "white-background" -> story should be called "white background"... and the likes.)

– code examples contain wrong code e.g.
<scale-toggle-group variant='secondary' disabled>

– icons-story: please don't mix (not a good example)
please add:
1. icon before 
2. icon only (include aria title on icon!!) 
"icon after" is not a case we'd recommend.

– not able to see the cases of e.g. xs with icons.

– story single select more prominent. (3rd place maybe?)

and little bits of design:

– design: disable border: color
Bildschirmfoto 2021-08-18 um 17 10 04

– check & adjust element roundings
Bildschirmfoto 2021-08-18 um 17 09 13

timheiler-ico:
toggle group needs a story featuring "Color Scheme Dark"

PR #514, opened 2021-08-10
chore: update Stencil to latest version

acstll:
@nowseemee do you remember any reason why we wouldn't update Stencil in the past? some SSR issue?


acstll:
Feel free to merge if there's nothing against it 🤓

PR #523, opened 2021-08-13
Feat/alertbox beta

Kutlovcidenis:
Adjusted Storybook like Tim reqeusted.


packages/components/src/components/alertbox/alertbox.css
```diff
@@ -0,0 +1,124 @@
...
+  color: var(--scl-color-primary-active);
+}
+
+/* Success Icon */
+.alertbox .alertbox__icon-success {
+  position: absolute;
+  top: 13px;
+  left: 17px;
+  height: 20px;
+  width: 20px;
```
acstll, line 85: There are many hard-coded values (numbers and colors) that could be tokens.

packages/components/src/components/alertbox/alertbox.tsx
```diff
@@ -0,0 +1,176 @@
...
+  // Color to variant rename
+  @Prop() variant?: 'informational' | 'success' | 'warning' | 'error' =
+    'informational';
+  @Prop({ reflect: true }) hasClose?: boolean = false;
+  @Prop({ reflect: true }) opened: boolean;
+  @Prop() timeout?: boolean | number = false;
+  /** (optional) aria-label attribute */
+  @Prop() ariaLabel: string;
+  /** (optional) aria-description attribute */
+  @Prop() ariaDescription: string;
```
acstll, line 28: This prop is unused, is that intentional?

packages/components/src/components/alertbox/alertbox.tsx
```diff
@@ -0,0 +1,176 @@
...
+      } else {
+        setTimeout(this.close, this.defaultTimeout);
+      }
+    } else {
+      return null;
+    }
+  };
+
+  render() {
+    this.onCloseAlertWithTimeout();
```
acstll, line 119: I would put this call in a lifecycle method, probably on [`componentDidLoad`](https://stenciljs.com/docs/component-lifecycle#componentdidload). It's a nice practice to avoid any logic inside the `render` function.

packages/components/src/components/alertbox/alertbox.tsx
```diff
@@ -0,0 +1,176 @@
...
+    if (!this.opened) {
+      return null;
+    }
+
+    return (
+      <Host>
+        <div
+          part={this.getBasePartMap()}
+          class={this.getCssClassMap()}
+          aria-label={this.ariaLabel}
```
acstll, line 130: This `ariaLabel` may be duplicated when set as an attribute in the host (`scale-alertbox`). You can end up with:

```html
<scale-alertbox aria-label="Foo">
  #shadow
    <div aria-label="Foo"></div>
</scale-alertbox>
```

Maybe this `ariaLabel` prop is not needed, I'm not sure what the right approach would be.

acstll (commented):
Thanks for updating the branch, it's much better now 👍

PR #532, opened 2021-08-19
feat: changes for adding list attribute to the scale text field.

gaurishkamat:


gaurishkamat:
Hi @acstll,
Created a new PR with suggested changes. Could you please have a look whenever you are free.

Thanks!

packages/components/src/components/text-field/text-field.spec.ts
```diff
@@ -198,4 +200,17 @@ describe('TextField', () => {
...
+    expect(textField.readData('no data')).toEqual([]);
+  });
+
+  it('should handle JSON error', () => {
+    const textField = new TextField();
+    expect(textField.readData(['arraydata', 'arraydata'])).toEqual([
+      'arraydata',
+      'arraydata',
+    ]);
+  });
```
acstll, line 215: This is not needed anymore, please remove.

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -100,6 +102,14 @@ export class TextField {
...
   // because how we keep this.value up-to-date for type="select"
   // `this.value = selectedValue`
+  readData = (data) => {
+    try {
+      return Array.isArray(JSON.parse(data)) ? JSON.parse(data) : [];
+    } catch (error) {
+      return Array.isArray(data) ? data : [];
+    }
+  };
+
```
acstll, line 112: Also not needed, please remove.

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -142,6 +152,7 @@ export class TextField {
       this.status === 'error' ? { 'aria-invalid': true } : {};
     const helperTextId = `helper-message-${i}`;
     const ariaDescribedByAttr = { 'aria-describedBy': helperTextId };
+    const list = this.list ? { 'list': this.list } : null;
```
acstll, line 155: We can inline the list attribute below, we can remove this.

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -160,6 +171,7 @@ export class TextField {
             minLength={this.minLength}
             maxLength={this.maxLength}
             id={this.inputId}
+            {...list}
```
acstll, line 174: This can be `list={this.list}`. If `this.list` is undefined, the attribute won't render.

acstll (changes requested):
There is some code we don't need and some that can be simplified, I left some comments.

Thank you for the quick response!

packages/components/src/components/text-field/text-field.tsx
gaurishkamat: @acstll done.

acstll:
@gaurishkamat some checks are not passing, could you be so kind and try and fix those?

- for the `prettier` check you can run `yarn format` (in the root or in packages/components)
- for the `e2e-tests` check make sure the e2e tests pass, run them with `yarn test --e2e` inside packages/components
- for the `uncommitted-changes` check, you need to run `yarn build` in packages/components (it will update the `text-field/readme.md` file automatically, those readme files are auto-generated)

Thanks again 🙏

gaurishkamat:
Thanks for the much needed detailed explanation @acstll. I have completed the above mentioned steps, let me know if anything is pending.

gaurishkamat:
@acstll not sure why the checks are failing.

acstll:
> @acstll not sure why the checks are failing.

me neither.

e2e pass locally for me. And `yarn build` generates a couple of readme.md files, but they're OK in `main` and your branch is up-to-date with `main`, so it's indeed weird.

Since e2e tests pass locally and the uncommitted changes seem to be OK too. I will merge.

Thanks a lot again for your contribution.

gaurishkamat:
Hi @acstll should this be working for `scale-components-react` as well or we need some more changes?

acstll:
It should, the React package is automatically generated. Does it work? the latest `3.0.0-beta.17` should have it.

PR #540, opened 2021-08-24
refactor(date-picker): a11y update

acstll:
- [x] remove Duet's `h2` with {Month Year} text, and link default `h2` with "Pick a date" text to dialog via `aria-describedby` 5f4c20d9e7df2b85b3df563b655f5caed6da3542
- [x] make `today` label in today button translatable ("Localization" story) 19f327d10a1de26a9cb1db4a6a4c49d45aebc90f
- [x] fix JAWS displaying the current day twice (e.g. "19 19 August, today") aa0aed25c412e8b32e79e7b2c1b380937803ab25


acstll:
Something I found on the Internet regarding `aria-hidden` being read by JAWS

> If you use `aria-describedby` in the same element as `aria-hidden="true"` the screen reader may read it.

> The `aria-hidden` attribute does not work on focusable elements such as form inputs, links, and buttons.

> If you use `aria-hidden` on an element with child elements, the child elements will be hidden as well.

AnnaKuchtin (approved):

acstll:
Sending it out… 🚀

PR #557, opened 2021-08-31
Refactor/size prop to string

ChrisPaj:
add prop 'size': string and make prop 'small': boolean deprecated;
Components:  tab-header, tav-navigation and tab-panel, pagination;
It should be possible to use both props afterwards, so no braking chains occur by using the deprecated 'small' prop;


packages/components/src/components/pagination/pagination.tsx
```diff
@@ -58,8 +58,10 @@ export class Pagination {
   @Prop() totalElements?: number = 1;
   /** (optional) Injected styles */
   @Prop() styles?: string;
-  /** (optional) small  */
-  @Prop() small = false;
+  /** DEPRECATED - size should replace small */
```
oddcelot, line 61: i can maybe cook up something for the deprecation notice, so you can pass in the props that are to be deprecated and when the value isn't nullish you get a warning in the console?
(or you can do it, depends on how much time you have 🙂)

oddcelot (approved):

oddcelot (approved):

PR #558, opened 2021-09-01
Refactor/rating stars adapt props

ChrisPaj:
change props 'starSize' to 'size' and 'maxRating' to 'max'; 
Props 'starSize',  'maxRating' and 'minRating' (will be dropped on next release) become deprecated;
To avoid breaking chains, the deprecated props are still valid, overwriting their successors;


packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -64,11 +68,22 @@ export class RatingStars {
...
+      this.size = this.starSize;
+    }
+  }
+
   // constructs the aria message for the current rating
   getRatingText() {
     const filledText = this.ariaLabelTranslation
       .replace(/\$rating/g, `${this.rating}`)
-      .replace(/\$maxRating/g, `${this.maxRating}`);
+      .replace(/\$maxRating/g, `${this.maxRating}`)
```
oddcelot, line 85: so when `this.maxRating` is not updated after init (see `connectedCallback()`) should we carry `maxRating` through the component?
like isn't this  out of sync somewhere? 😅

oddcelot (commented):
we may want to make a `console.warn()` announcement when someone uses the deprecated props 🤔

packages/components/src/components/rating-stars/rating-stars.tsx
```diff
@@ -64,11 +68,22 @@ export class RatingStars {
   /** @deprecated in v3 in favor of kebab-case event names */
   @Event({ eventName: 'scaleChange' }) scaleChangeLegacy: EventEmitter;
 
+  connectedCallback() {
+    // make sure the deprecated props overwrite the actual ones if used
+    if (this.host.hasAttribute('max-rating')) {
```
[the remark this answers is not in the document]
ChrisPaj: The idea is, if someone uses a max-rating attribute of e.g. 10, the default max prop has to be overwritten. If you checked for the prop maxRating, you always would overwrite max with maxRating's default value. It is a good idea not to reflect the deprecated props any more.

packages/components/src/components/rating-stars/rating-stars.tsx
ChrisPaj: this one only gets effective, if the user uses a customised aria-label-translation attribute containing '$maxRating'

PR #588, opened 2021-09-20
Dx/usage assets

oddcelot:


oddcelot:
this makes the usage docs for the components a little easier to edit and preview (also directly in github 😉)

take a look at https://github.com/telekom/scale/blob/dx/usage-assets/packages/storybook-vue/stories/3_components/button/button.md

PR #591, opened 2021-09-22
Refactor/flyout menu a11y

acstll:
- [ ] complete item styles (proper focus styles)
- [ ] complete list styles
- [ ] enable scrolling with indicators
- [ ] replace original component
- [ ] tidy up comments and class method/props order

🤘

### DIFF
--- a/packages/components-angular/src/directives/value-accessor.ts
+++ b/packages/components-angular/src/directives/value-accessor.ts
@@ -1,7 +1,6 @@
-import { Directive, ElementRef, HostListener } from '@angular/core';
+import { ElementRef, HostListener } from '@angular/core';
 import { ControlValueAccessor } from '@angular/forms';
 
-@Directive({})
 export class ValueAccessor implements ControlValueAccessor {
 
   private onChange: (value: any) => void = () => {/**/};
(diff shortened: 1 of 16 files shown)


acstll:
> this is quite radical - what if the consumer wants the trigger to be a web component or an anchor tag?

@nowseemee I have to admit is a bit radical, but let me explain:

https://github.com/telekom/scale/pull/591/commits/7d09fc4a4aad7bfeb924b3992f32bb1ce928ba4d#diff-b8db44c3fa5edf93b68e59c5864cc400c5b6ae021ead3fced77cd12e6d34e3efR93-R102

this only affects setting of the `aria-haspopup` attribute and setting and toggling `aria-expanded`, but the menu will continue to work even if they don't use a button (I tested the brand header examples)

if they don't use a button they loose keyboard support (opening the menu with Enter key), and thus also the corresponding attributes being set correctly —to me that's an OK trade-off

what I would do, following your suggestion, would be to also to check for `role=button` and `a` tags (in practice, you want to trigger the menu with either a `button`, or `div[role=button]`, or `a`; and if you use a web component, it will look for the same things inside the shadow root

so these elements will be considered as triggers:
- `button`
- `*[role="button"]` (no keyboard support by default)
- `a` (some keyboard support)
- any of those three inside the shadowRoot, if present

whatcha think?

PR #607, opened 2021-10-05
Feat/tooltip

marvinLaubenstein:


packages/components/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,164 @@
+host*,
```
acstll, line 1: Is this a valid selector?

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,164 @@
...
+  display: contents;
+  position: relative;
+  box-sizing: border-box;
+}
+
+.tooltip-positioner {
+  position: absolute;
+}
+
+#slot-container {
```
acstll, line 35: Is there a reason for this to be an id and not use a class? In the .tsx file it could use a `ref` instead of querying the DOM I think…

packages/storybook-vue/stories/3_components/tooltip/Tooltip.stories.mdx
```diff
@@ -0,0 +1,291 @@
...
+    {SlotTemplate.bind({})}
+  </Story>
+</Canvas>
+
+```html
+<scale-tooltip>
+  <p>
+    I'm a HTML-tooltip
+    <a slot="content" href="https://github.com/telekom/scale"> with a link</a>
+  </p>
```
acstll, line 268: I think this needs to be:

```html
<scale-tooltip>  
  <a slot="content" href="https://github.com/telekom/scale">I'm a HTML-tooltip with a link</a>
  <scale-button>Hover me</scale-button>
</scale-tooltip>
```

to match the actual template.

acstll (changes requested):
It's amazing to see this finished! Thank you @marvinLaubenstein ❤️

packages/components/src/components/tooltip/tooltip.css
marvinLaubenstein: Yes I think it was something like selecting all child elements of the parent element(host)

packages/storybook-vue/stories/3_components/tooltip/Tooltip.stories.mdx
marvinLaubenstein: Oh yes, you are right. Thank you. Changed

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,164 @@
...
+  --scale-tooltip-arrow-size: 5px;
+  --scale-tooltip-arrow-start-end-offset: var(--scl-spacing-8);
+
+  --scale-tooltip-max-width: 20rem;
+  --scale-tooltip-hide-delay: 0s;
+  --scale-tooltip-hide-duration: 0.125s;
+  --scale-tooltip-hide-timing-function: ease;
+  --scale-tooltip-show-duration: 0.125s;
+  --scale-tooltip-show-timing-function: ease;
+  --scale-tooltip-z-index: var(--scl-z-index-70);
```
acstll, line 24: We need to rename these following the conventions 🤓 

https://gist.github.com/acstll/13b005c6c75e0e550c656b3fdaecb965#naming-convention-for-component-scoped-css-variables

packages/components/src/components/tooltip/tooltip.css
acstll: I tested it and it seems `host*` is not selecting any elements 🤪 (you can see that in devtools)

You can either use:
- `:host` for selecting the host, in this case `scale-tooltip`
- `:host(selector)` to selecting the `scale-tooltip` with some extra selectors like a class, or
- `*` to select all elements inside the shadow root, _excluding_ the host

Is the `box-sizing: inherit;` declaration actually needed? if so, in which elements?

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,163 @@
+host*,
+*::before,
+*::after {
+  box-sizing: inherit;
+}
+
+:host {
+  /*tooltip*/
+  --border-radius: var(--scl-radius-4);
```
acstll, line 9: can we use `--radius`?

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,163 @@
...
+  --background-color: var(--scl-color-grey-60);
```
acstll, line 10: can we use `--background`?

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,163 @@
...
+  --color: var(--scl-color-white);
+  --font-weight: var(--scl-font-weight-regular);
+  --font-size: var(--scl-font-size-16);
+  --line-height: 1.4;
+  --padding: var(--scl-spacing-2) var(--scl-spacing-8);
```
acstll, line 15: can we use `--spacing`?

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -0,0 +1,163 @@
...
+  /*arrow*/
+  --arrow-size: 5px;
+  --arrow-start-end-offset: var(--scl-spacing-8);
```
acstll, line 18: I would use `--arrow-offset` ("start-end" seems confusing to me)

acstll (commented):
Thanks for the update! I'm asking for a couple of changes in the names (sorry for being so picky, but this is hard to change later on)

Also, the selector issue is still pending.

Thank you 🙏

packages/components/src/components/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  @Watch('open')
+  handleOpenChange() {
+    this.open ? this.showTooltip() : this.hideTooltip();
+  }
+
+  @Event({ eventName: 'scale-show' }) tooltipShow: EventEmitter;
+  @Event({ eventName: 'scale-aftershow' }) tooltipAfterShow: EventEmitter;
+  @Event({ eventName: 'scale-hide' }) tooltipHide: EventEmitter;
+  @Event({ eventName: 'scale-after-hide' }) tooltipAfterHide: EventEmitter;
```
acstll, line 73: To follow the same pattern we have in the Modal, I would rename these events:

- `scale-show` → `scale-before-show`
- `scale-aftershow` → `scale-show`
- `scale-hide` → `scale-before-hide`
- `scale-after-hide` → `scale-hide`

This way to can `preventDefaultp the action with the "before" event, and you get the final "show" or "hide" events after they actually happen. Does that make sense to you?

packages/components/src/components/tooltip/tooltip.css
marvinLaubenstein: Wouldn't it be better to keep padding, due to consistency ? Because we use for all other variables names exactly the name for the matching property

acstll: Yes, but we also use `--spacing` in almost all the other components. Maybe it wasn't the best decision using `spacing` in the first place, but it's only because of consistency within the library that I'm suggesting to use `--spacing`.

The reason for using `spacing` is to keep it independent from the internal implementation, so to avoid having something like `margin: var(--padding)` if you ever need to refactor the component and use margins instead of padding, or viceversa.

All other suggestions follow this same principle, consistency across components. Thanks for asking 🙏

packages/components/src/components/tooltip/tooltip.css
marvinLaubenstein: I needed it before but it seems to do nothing in the actual version, so I removed it for now.😅

Thanks for the detailed explanation😁

PR #641, opened 2021-10-25
feat: add overview of changed props CHANGELOG

ChrisPaj:


acstll (commented):
Thank you @ChrisPaj this is a great idea!

I'm not sure the CHANGELOG.md file is the best place for it (though it makes sense), because it's a generated file and there might be conflicts. (What do you think @nowseemee ?).

I would create a new page in Storybook under "Scale for Developers" with the name "Deprecation Notices"… and then we can add other stuff in there, like this https://github.com/telekom/scale/blob/main/packages/storybook-vue/stories/3_components/flyout-menu/MenuFlyout.stories.mdx#breaking-changes-from-300-beta22

Then, I would suggest the following changes for the table itself, you tell me what you think:
- use four columns: component, deprecated prop, new prop, PR (so we don't have empty cells that look weird and the info is more linear)
- wrap the props in backticks so it's rendered with a mono font (probably easier to read)
- if possible, add a fifth column called "from version" with the version number in which the changed was released
- maybe: change the title from "Harmonisation of props" (which otherwise sounds lovely) to something like "Deprecated props", because this harmonisation idea is something we know and have talked about and refers to the internal process taking place, but can sound strange to developers just looking for deprecated props.

Thanks again! 🙏

PR #651, opened 2021-10-27
rename badge component to callout

ChrisPaj:


acstll (changes requested):
Only 2 minor comments for the renaming:
- I found a bypassed visual test file there `packages/visual-tests/src/badge.visual.spec.js.x` — but maybe this has been already addressed in another PR?
- I think the prettier check failing is relevant (something tiny in the `index.html`)

Other than that, this is good to go 👍 👍 

---
Nevertheless, I think this component can be further improved in two areas:
- using more slots or slots in a smarter way (right now there's too much user-provided markup and styles required)
- renaming the CSS scoped variables following the conventions*

But this could be done in another PR and keep this one clean as it is, do you agree?

*_which we need to document somewhere (we had this on some PR, maybe in GitLab haha), so could be a good time to do it_

acstll (approved):

PR #662, opened 2021-11-01
prop harmonisation replace prop by css

ChrisPaj:
progress bar and slider: replace prop 'customColor' by css variable

PR #663, opened 2021-11-01
refactor(toggle-group): replace prop 'colorScheme' by 'variant'

ChrisPaj:


acstll (changes requested):
Same as in #662, I would use the JSDoc `@deprecated` syntax instead of `DEPRECATED` 🙂

PR #665, opened 2021-11-01
prop harmonisation replace status prop

ChrisPaj:
- replace prop 'status' (String) by 'invalid' (Boolean)
- avoid breaking changes


packages/components/src/components/date-picker/date-picker.tsx
```diff
@@ -22,6 +22,7 @@ import {
   Host,
 } from '@stencil/core';
 import { DuetDatePicker as DuetDatePickerCustomElement } from '@duetds/date-picker/custom-element';
+import statusNote from '../../utils/status-note';
```
acstll, line 25: This is import is duplicated below so Stencil fails to build 👀

acstll (changes requested):
Same suggestion from https://github.com/telekom/scale/pull/662/files#r740799313 And that duplicated import.

Otherwise it's 👌, great work @ChrisPaj — Thank you!

acstll (approved):

PR #623, opened 2021-10-14
first implementation of grid

halfas:
Fisrt implementation of layout based on css grid.

PR #674, opened 2021-11-04
feat: grid beta (RFC)

acstll:
This is a simpler take based on the work on #623 

We need your feedback!

I will post a draft of the documentation, and then a few questions. Please make sure you read all the way through 🤓 

---

# Draft docs for `scale-grid`

The spec (as in Storybook):

| breakpoint | suffix | columns available | margin | gutter |
| ------- | --- | ------- | ------- | ------ |
| default | sm  | 4       | 16px    | 32px   |
| >640px  | md  | 8       | 16px    | 32px   |
| >1040px | lg  | 16      | 24px    | 32px   |
| >1296px | xl  | 16      | 24px    | 32px   |
| >1552px | xxl | 16      | 24px    | 32px   |

## API

- There are 2 custom elements: `scale-grid` and `scale-grid-item`.
- At the moment, these aren't real "web components" —only custom elements—, so it works without JavaScript.
- There's also CSS classes: `.scl-grid` and `.scl-grid-item`, that do exactly the same thing.
- Settings are all controlled via CSS variables.
- Each variable has a breakpoint suffix: `--size-sm`, `--size-md`, etc.

### CSS variables for `scale-grid` or `.scl-grid`

- **gutter-y:** `--gutter-y-{suffix}` - space between columns (`32px`)
- **gutter-x:** `--gutter-x-{suffix}` - space between rows (`32px`)
- **spacing:** `--spacing-{suffix}` - outer padding (`16px`, `24px`)
- **max-width:** `--max-width` - max width (`1552px`, not responsive)

### CSS variables for `scale-grid-item` or `.scl-grid-item`

- **size:** `--size-{suffix}` - number of columns the item spans (`4`)
- **offset:** `--offset-{suffix}` - where the item is placed, from the left (`auto`)

### Usage

With default size value of 4 columns.

```html
<scale-grid>
  <scale-grid-item>
    A
  </scale-grid-item>
  <scale-grid-item>
    B
  </scale-grid-item>
  <scale-grid-item>
    C
  </scale-grid-item>
  <scale-grid-item>
    D
  </scale-grid-item>
</scale-grid>
```

Or with CSS classes only (outcome is the same):

```html
<div class="scl-grid-item">
  <div>
    E
  </div>
  <div class="scl-grid-item">
    F
  </div>
  <div class="scl-grid-item">
    G
  </div>
  <div class="scl-grid-item">
    H
  </div>
</div>
```

Different column sizes across breakpoints (remember default size in every breakpoint is `4`):

```html
<scale-grid>
  <scale-grid-item style="">
    letterpress
  </scale-grid-item>
  <scale-grid-item style="
    --size-lg: 12;
    --size-xl: 12;
    --size-xxl: 12">
    disrupt
  </scale-grid-item>
  <scale-grid-item style="
    --size-lg: 8;
    --size-xl: 8;
    --size-xxl: 8">
    unicorn
  </scale-grid-item>
  <scale-grid-item style="
    --size-lg: 8;
    --size-xl: 8;
    --size-xxl: 8">
    semiotics
  </scale-grid-item>
  <scale-grid-item style="
    --size-lg: 12;
    --size-xl: 12;
    --size-xxl: 12">
    readymade
  </scale-grid-item>
  <scale-grid-item style="">
    hashtag
  </scale-grid-item>
</scale-grid>
```

Captura de Pantalla 2021-11-04 a les 14 35 16

---

# And now the topics where we need your feedback!

## (a) should larger breakpoints inherit the value of the previous ones?

currently if you want to change the size of an item, you have to specify the size for EVERY breakpoint, e.g. let's say you want to have a 2-column layout, you would have to set the sizes like this:
- sm: 2 (out of 4)
- md: 4 (default) (out of 8)
- lg: 8 (out of 16)
- xl: 8
- xxl: 8

some libraries I like (https://tachyons.io/) leverage the cascade, so you would have enough with this:

- sm: 2
- md: 4
- lg: 8

because `xl` and `xxl` inherit the value from `lg`.

I prefer the "cascading" approach. Do you?

## (b) is using "breakpoint variables" (with the breakpoint suffix) a bad idea?

in general it's a stupid idea (`--size-xl`) —you would normally change the value of the same variable (`--size`) across breakpoints—, but here it kinda makes sense?

## (c) different number of columns per breakpoints is problematic

having the underlying grid have 4 columns first, then 8 and then 16, makes it really hard for developers to use it, because in order to set the `size` you have to know how much columns you have available (compared for instance to being able to say "I want 2 columns" and go: 50% + 50% for percentages, or 6 + 6 for the quite common 12-column systems)

you're not going to remember the number of columns per breakpoint, neither how many breakpoints there are

extra point: "3 thirds" layouts (33% 33% 33%) are not possible

## (d) the CSS variables API is annoying?

in the beginning it felt like a good idea, but after testing it I didn't quite like it — we could turn the custom elements into real web components and have a different API with attributes that simply set the CSS variables under the hood (as in the original #623 proposal)

## (e) the system is not super flexible

you cannot build all kinds of layouts with this, it's meant to follow the spec, so I guess that's OK?

## (f) should `max-width` be responsive?

currently you can only set it once, for the largest breakpoint (I don't see the point in limiting the width of the layout below that, if not for having the column sizes be multiples of 8px)


reld:
This looks very cool ⚡ nice work!

Now to your thoughts:
- A) Definitely the cascade approach. I makes things less complex and more clean in my opinion.
- B) In this case I think it makes for the sake of clarity and orientation when using the component. Although it might feel stupid from the coding point of view.
- C) Maybe we can make a wrapper component where we can pass the number of columns as a prop and the  number of breakpoints and have this components handle the tricky stuff?
- D) Here it definitely make sense to me to create real components to help us with the abstraction and make it more dev friendly.
- E) Here I would like us to make a proposal to make a robust "layout system" where we can create any kind of layouts a' la everylayout per instance.
- F) I don't understand this one 😅

Cheers 🤙

acstll:
> What happened to this feature?

@filipjnc it's coming soon! A first version of the component is already in the repo, only the documentation is missing.

PR #678, opened 2021-11-05
feat: add tabIndex prop to button and link

halfas:


packages/components/src/components/link/link.tsx
```diff
@@ -78,7 +80,7 @@ export class Link {
         }}
       >
         {this.styles && <style>{this.styles}</style>}
-        <a part="anchor" {...this.getAnchorProps()}>
+        <a part="anchor" {...this.getAnchorProps()} tabIndex={this.innerTabindex}>
```
acstll, line 83: I think you need to add the attribute in the `getAnchorProps` function, [see line 60](https://github.com/telekom/scale/blob/b564877ca6fadc2451986a3b09f4a9a83fd5b333/packages/components/src/components/link/link.tsx#L60)

I think this should work:

```js
tabIndex: this.disabled ? -1 : this.innerTabindex
```

acstll (changes requested):

acstll (approved):

PR #684, opened 2021-11-08
refactor: hover background

marvinLaubenstein:
little refactor
Bildschirmfoto 2021-11-08 um 21 15 56
h


packages/components/src/components/tag/tag.css
```diff
@@ -92,7 +92,7 @@
   box-shadow: var(--box-shadow-button-dismissable-focus);
 }
 .tag--dismissable button:hover {
-  background: var(--background-button-dismissable-hover);
+  background: var(--scl-color-white);
```
acstll, line 95: Did you consider setting `--background-button-dismissable-hover` in the root to `var(--scl-color-white)` instead of leaving it empty?

acstll (commented):
Thank you @marvinLaubenstein ! This is long overdue :)

packages/components/src/components/tag/tag.css
marvinLaubenstein: Yes that make definitely sense. However, I actually did not notice it at all in the root. Otherwise I would have changed it there. Thanks 😁

marvinLaubenstein: Yes that make definitely sense. However, I actually did not notice it at all in the root. Otherwise I would have changed it there. Thanks 😁

PR #704, opened 2021-11-24
Feat/notification banner

marvinLaubenstein:


packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -0,0 +1,177 @@
...
+  shadow: true,
+})
+export class NotificationMessage {
+  @Element() hostElement: HTMLElement;
+
+  @Prop() variant?: 'informational' | 'success' | 'warning' | 'error' =
+    'informational';
+  @Prop() dismissible?: boolean = false;
+  @Prop({ reflect: true }) opened: boolean;
+  @Prop() timeout?: boolean | number = false;
```
nowseemee, line 38: this is confusing - can we think of a better name? like auto-hide? and auto-hide-duration? so we don't have two types for one prop and the naming is self explanatory.

packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -0,0 +1,177 @@
...
+        case 'error':
+          return (
+            <scale-icon-alert-warning
+              class="notification-banner__icon-error"
+              accessibility-title="error"
+            />
+          );
+        case 'warning':
+          return (
+            <scale-icon-alert-information
```
nowseemee, line 94: so we have the information icon 2x?

packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -0,0 +1,177 @@
...
+    if (!this.opened) {
+      return null;
+    }
+
+    return (
+      <Host>
+        <div
+          part={this.getBasePartMap()}
+          class={this.getCssClassMap()}
+          tabindex="0"
```
nowseemee, line 119: does it really need the tabindex? What is it good for? If I'd have to come up with an answer, I'd say we should make this element focused for screenreader when it appears - did you think about this?

packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -0,0 +1,177 @@
...
+        >
+          <div part="container" class="notification-banner__container">
+            {this.handleIcons()}
+            <div part="heading" class="notification-banner__heading">
+              <slot>&emsp;</slot>
```
nowseemee, line 124: 4 non-breaking spaces? what is this good for?

packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+              {this.dismissible && (
+                <scale-icon-action-circle-close
+                  tabindex="0"
```
nowseemee, line 128: what if we wrap the icon into a button? would that improve accessibility? and it would take away the need of this tabindex.

packages/components/src/components/notification-banner/notification-banner.css
```diff
@@ -0,0 +1,122 @@
...
+.notification-banner__text ::slotted(*) {
+  margin: -25px 0px -25px 0px;
+  padding: 0;
+}
+
+.notification-banner--has-text.notification-banner {
+  margin-top: -25px;
+}
+
+.notification-banner--has-no-text .notification-banner__link ::slotted(*) {
```
nowseemee, line 58: this is not used, right?

packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -0,0 +1,177 @@
...
+                  }}
+                  onKeyDown={(e) => {
+                    if (e.key === 'Enter') {
+                      this.close();
+                    }
+                  }}
+                  accessibility-title="close"
+                />
+              )}
+              {this.hasText ? (
```
nowseemee, line 141: can you use https://developer.mozilla.org/en-US/docs/Web/CSS/:empty instead of javascript to determine if the box needs spacing or not?

nowseemee (commented):
this notification-banner feels like a duplication of notification-message - can we enhance the original component with the features this new one is introducing? or if we insist on having 2 separate components then we could reuse notification-message inside the banner. WDYT?

packages/components/src/components/notification-banner/notification-banner-svg.tsx
```diff
@@ -0,0 +1,62 @@
+/**
```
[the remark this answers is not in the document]
marvinLaubenstein: Theoretically I can kick it out the and use the already existing notification message instead, then we would not have a duplicate and would use it in 3 places

packages/components/src/components/notification-banner/notification-banner.css
```diff
@@ -0,0 +1,122 @@
...
+  --background-warning: hsla(33.5, 100%, 75.1%, 0.3);
+  --background-informational: hsla(190.20000000000005, 100%, 92%, 0.75);
+  --background-success: hsl(98.80000000000001, 49.7%, 61%, 0.2);
+}
+
+.notification-banner {
+  border-radius: var(--radius);
+  width: var(--width);
+  position: relative;
+  box-shadow: 0px 0px 10px #e2e2e2;
```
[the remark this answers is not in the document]
marvinLaubenstein: Yes of course, I am so sorry, I forgot the token thing

packages/components/src/components/notification-banner/notification-banner.css
```diff
@@ -0,0 +1,122 @@
...
+}
+
+.notification-banner__icon-close:hover svg {
+  background-color: white;
+  color: var(--scl-color-primary-hover);
+}
+
+.notification-banner__icon-success {
+  position: absolute;
+  top: 14px;
```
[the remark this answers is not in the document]
marvinLaubenstein: This is due to the success icon which does not come from the icons, but from the additional component

packages/components/src/components/notification-banner/notification-banner.css
```diff
@@ -0,0 +1,130 @@
...
+}
+
+.notification-banner__text ::slotted(*) {
+  margin: -10px 0px -25px 0px;
+  line-height: 16px;
+  padding-bottom: 15px;
+}
+
+.notification-banner--has-text.notification-banner {
+  margin-top: -25px;
```
nowseemee, line 57: do you have a reason for having -25px? can you remove this all together?

packages/components/src/components/notification-banner/notification-banner.css
nowseemee: you resolved this question without answering - no problem, but i have to ask again: notification-banner--has-no-text is nowhere to be found but here. can you remove this then?

nowseemee:
@marvinLaubenstein could you fix the conflicts and update the branch with the latest main? thanks! :)

PR #713, opened 2021-12-01
Refactor/notification message

marvinLaubenstein:
Design changes for the icons

Prop naming change


acstll (approved):
Thank you @marvinLaubenstein this is great!

It'd be cool to use tokens for all colors and current values using hard-coded pixels (for like positioning the icons). If there are no tokens for the values needed, we can talk to designers to either change the design a bit or see if it makes sense adding new tokens (never adding tokens without previous consensus though).

Also, the colors for icons being hard-coded in the markup makes it impossible (I think) for people to customise them. We need to change this.
 
Since this is officially beta, we'll have time to do this improvements in new PRs :)

PR #716, opened 2021-12-02
feat: add prop alwaysScrolled to header

mato-a:
# Add property alwaysScrolled to header

If this prop is set to true header will be always thin even if the page was not scrolled down.

Class `header--sticky` will be added always to the header.

This is useful for example in fullscreen mode where I'd like to make the navbar as small as possible.

### DIFF
--- a/packages/components/src/components/telekom/app-header/app-header.tsx
+++ b/packages/components/src/components/telekom/app-header/app-header.tsx
@@ -57,7 +57,7 @@ export class Header {
   @Prop() addonNavigation?: any = [];
   @Prop() activeRouteId: string;
   @Prop() activeSectorId?: string;
-  @Prop() alwaysScrolled?: boolean = false;
+  @Prop() sticky?: boolean = false;
   // DEPRECATED - megaMenuVisible should replace isMegaMenuVisible
   @Prop() isMegaMenuVisible?: boolean = false;
   @Prop() megaMenuVisible?: boolean = false;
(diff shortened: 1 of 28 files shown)


nowseemee:
Thanks for your PR! 🎉 🎉 🎉 
I'll discuss this with the team and come back to you soon.

packages/components/src/components/telekom/app-header/app-header.tsx
```diff
@@ -57,6 +57,7 @@ export class Header {
   @Prop() addonNavigation?: any = [];
   @Prop() activeRouteId: string;
   @Prop() activeSectorId?: string;
+  @Prop() alwaysScrolled?: boolean = false;
```
nowseemee, line 60: can you rename the new prop to `sticky` please?

nowseemee (changes requested):
Thanks a lot for your PR! I'd like you to change the name of the prop to `sticky` and also add this to `app-shell` and pass it down to `app-header`.

nowseemee:
Thanks for adding your changes. Can you please run `yarn build` in the `components` package and commit the changes? Many thanks for all your contributions!

nowseemee:
thanks a lot @mato-a - i'll publish this to NPM today and let you know ✌🏻

mato-a:
@nowseemee great 👍 

Thanks for guidance 😉

nowseemee:
@mato-a, it's released: `3.0.0-beta.33`

PR #725, opened 2021-12-09
feat: adapt icon size automatically based on context

### LINKED ISSUE #702: Inherit icon size to font size of text in button
Hi, 

I have something like this : 
```
<scale-button
    variant="ghost" 
    type="button"
    icon-position="after"
    size="small"
  >
   button-label
    <scale-icon-user-file-attachment size="16"></scale-icon-user-file-attachment>
  </scale-button>
```
Here in the code if I remove size="16" from the icon, the icon is much higher than the text in button (button-label). Is it somehow possible make it "dynamic" if button use size=small than also icon should inherit that ? or at least use the same "naming" for size="small" also in icon ? not use direct numbers? It is useful if I need to have larger icon to set its size with a number but not all the time.

Maybe it is actually somehow possible, but my question related to the usage of button with icon, if button has attribute `size="small"` than I have to manually do set `<icon-XX size="16"/>` .

 One possible solution should be to allow for size attribute in icon `types: number | 'small' | 'large'` as allowed types similar also to the size attribute used in button, and in that case, the "font-size" should use same size for button text and icon also ? Or maybe I am just missing some "docs" how to correctly use that ?

acstll:
Fixes #702

- [x] button
- [x] tab nav
- [ ] ~~toggle button group~~ (different PR)
- [x] ~~unit~~ e2e tests


reld (approved):
@acstll Couldn't find anything funny! :)

PR #772, opened 2022-01-07
fix: add flex value to main content

maomaoZH:
Set main content flex value to 1.


nowseemee (approved):
thanks for the PR! 🎉 🎉 🎉

PR #787, opened 2022-01-12
fix: set select box default background color

maomaoZH:


nowseemee:
hi @maomaoZH thanks for the PR! 🎉  🎉  🎉 
can you please update the visual snapshots of the dropdown? thanks!

PR #807, opened 2022-01-20
fix: spacing control is never applied on checkbox

Lalaluka:
The CSS Variable `--spacing-control` on the Checkbox is never applied instead it directly applies a design token variable.
I think this was the intended way the variable should work.


nowseemee:
Thanks for the PR! 🎉  🎉  🎉

PR #808, opened 2022-01-21
Feat/issue 801 email type

maomaoZH:


nowseemee:
Hi @maomaoZH, thanks for the PR! 🎉 🎉 🎉 

Could you add the new stories to the visual test suite and capture a new snapshot of those, please?

This is the file, where you should list the new stories along the existing one in the test.each: `scale/packages/visual-tests/src/datagrid.visual.spec.js`

Issue #818, opened 2022-01-28
Tooltips does not overlay elements


tommy-codez:
Hi there,

it seems the Tooltip doesnt overlay elements in a page. Sometimes the label/fields are showing and the tooltip is underneath.
Also there are some issues when the tooltip is on an edge of a site. This behavior differs in different browsers..

I made a small example of this... https://codesandbox.io/s/tooltip-y4y5k

Greetings,

Lalaluka:
Took a look.
The z-index seems to never be applied.
https://github.com/telekom/scale/blob/8545c211438ee21d89191181ab8314bec7eeaf8f/packages/components/src/components/tooltip/tooltip.css#L32

Adding the z-index manually to the shadow dom seems to fix it.
I might try to write a fix later :)

Lalaluka:
Fixing it is pretty trivial (just one line of CSS) see: #821

maomaoZH:
Dear @tommy-codez,  Thanks for reporting this issue. 
@Lalaluka Thanks for the fix!

nowseemee:
Meanwhile, this patch got published: `3.0.0-beta.46` - Huge thanks to @Lalaluka!!! 🚀

PR #821, opened 2022-01-29
fix(tooltip): add z-index

Lalaluka:
The css Variable of the Tooltips adding the z-index is never applied. This leads to behavior described in #818

This fixes this applying it to `tooltip-positioner`.
Can be tested by applying this to the index.html:
```
<scale-tooltip
      content="Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lore"
      title="Some"
  >
  <scale-icon-alert-information size="18" />
</scale-tooltip>
<scale-text-field
  size="small"
  label="Textfield"
  name="Name"
  id="FieldID"
  value="some"
/>
```

### DIFF
--- a/packages/components/src/components/tooltip/tooltip.css
+++ b/packages/components/src/components/tooltip/tooltip.css
@@ -38,6 +38,7 @@
 
 .tooltip-positioner {
   position: absolute;
+  z-index: var(--z-index);
 }
 
 .slot-container {


nowseemee:
Thanks for the PR! 🎉 🎉 🎉

PR #836, opened 2022-02-01
fix(checkbox): do not emit change from watching `disabled`

### LINKED ISSUE #817: React - ScaleCheckbox: disabled triggers onScaleChange
we have a `ScaleCheckbox` that gets `disabled` after click. 
we do a request and enable the checkbox when we get the response.
the seconed time we switch disabled state,  `onScaleChange` is called and unchecks the `ScaleCheckbox`

with `onChange` it works

im not a stencil expert, but could the `emitEvent` in `handleDisabledChange` be the Problem?

https://github.com/telekom/scale/blob/8545c211438ee21d89191181ab8314bec7eeaf8f/packages/components/src/components/checkbox/checkbox.tsx#L88-L93

acstll:
Fixes #817 

(Emitting `scaleChange` when `disabled` changes doesn't make much sense. I guess it was there to support the Checkbox Group?)

### DIFF
--- a/packages/components/src/components/checkbox/checkbox.tsx
+++ b/packages/components/src/components/checkbox/checkbox.tsx
@@ -17,7 +17,6 @@ import {
   h,
   Host,
   Prop,
-  Watch,
 } from '@stencil/core';
 import { emitEvent } from '../../utils/utils';
 import statusNote from '../../utils/status-note';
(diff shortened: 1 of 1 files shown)


nowseemee (approved):
👍🏻

PR #856, opened 2022-02-08
Refactor/semantic tokens textarea label

maomaoZH:


packages/components/src/components/textarea/textarea.tsx
```diff
@@ -175,36 +175,41 @@ export class Textarea {
...
-            {...(!!this.rows ? { rows: this.rows } : {})}
-            {...(!!this.cols ? { cols: this.cols } : {})}
-            {...ariaInvalidAttr}
-            {...(this.helperText ? ariaDescribedByAttr : {})}
-          />
-
-          {/* Accessibility: solid background for the textarea label to avoid making the label unreadable when there's text underneath */}
-          <span class="textarea__label-safety-background" aria-hidden="true" />
+          <div
+            class="textarea-mask"
```
felix-ico, line 179: I'm not sure if this follows convention or if should be `textarea__mask` instead?

packages/components/src/components/textarea/textarea.css
```diff
@@ -38,11 +38,11 @@ scale-textarea {
   --color-meta-error: var(--telekom-color-text-and-icon-functional-danger);
 
   /*control*/
-  --spacing-control: var(--telekom-spacing-unit-x6) var(--spacing-x)
-    var(--telekom-spacing-unit-x3) calc(var(--spacing-x) - 1px);
+  --spacing-control: 0 var(--spacing-x) var(--spacing-x);
   --transition-control: var(--transition);
   --font-size-control: var(--telekom-typography-font-size-body);
   --background-control: var(--telekom-color-ui-base);
+  --margin-control: var(--spacing-x);
```
felix-ico, line 45: I noticed that this --margin-control is used for vertical positioning of the label, perhaps we should use `--margin-control: var(--telekom-spacing-unit-x3);` directly to avoid confusion?

maomaoZH:
My latest push has updated such:
1 rename `textarea-mask` to `textarea__wrapper`;
2 set label position back to absolute and keep label transform value as it was
3 rename css variable `--spacing-x` to `--spacing-x-control` and add two more variables `--spacing-bottom-control` and `--spacing-top-control`
4 register click event handler to `textarea__wrapper` which will set textarea focus when clicking wrapper element

acstll (approved):

PR #862, opened 2022-02-09
fix: add part attributes to tooltip, follow BEM

felix-ico:


felix-ico:
closing this one, it's stale

Issue #806, opened 2022-01-20
Feature Request: Option to choose location of icon in Accordion


Lalaluka:
Currently, the only possible position of the Icon on the Collapsibles is left. In some designs, a location on the right would be more convenient. for example like this (ignore the borders):
<img width="382" alt="Bildschirmfoto 2022-01-20 um 19 13 34" src="https://user-images.githubusercontent.com/26364692/150397294-1516f566-6e4c-4eaa-a4f7-18ac441082f5.png">

Could such a new attribute be supported? For example called `iconLocation`?

maomaoZH:
Dear @Lalaluka 

Thanks for reaching us. It is a good idea to give user a flexibility to choose. We will get back to you asap. 

Cheers,
Maomao

Lalaluka:
Any news on this? 

If you currently don't have the capabilities to develop the feature I could write a PR myself.

maomaoZH:
Hi @Lalaluka,  we are currently indeed a bit overloaded. Please feel free to create a PR. We will be very happy to review it.

Lalaluka:
Hi @maomaoZH 
I created #863 :)

Lalaluka:
Hi @maomaoZH,
could you take a look at the PR? I kind of bricked my fork 😅 So merging it would enable me to renew my fork

maomaoZH:
@Lalaluka The new release `3.0.0-beta.51` is done. This includes your PR. Thanks again for your contribution!

PR #863, opened 2022-02-09
Feat: Accordion option for the icon being displayed on the right

Lalaluka:
This PR introduces a new attribute to accordions/collapsibles called `iconLocation`. This can be either set to `left` (default) or `right` changing the location of the icon of the collapsible as mentioned in #806

I also updated the storybook and sketch generator accordingly.


maomaoZH:
LGTM.  @nowseemee Can you please have a look also?

nowseemee (approved):
thanks for the PR! 🎉 🎉 🎉

VictoriaStefanikova:
Hi! I would like to ask if this feature is already usable. I tried to add iconLocation="right" attribute to my accordion, but it somehow doesn´t affect anything. Can you please suggest what am I doing wrong? Thank you.

```
<div>
  <scale-accordion expanded iconLocation="right">
    <scale-collapsible>
      <h2>General Information</h2>
      <p>some paragraph text</p>
    </scale-collapsible>
  </scale-accordion>
</div>
```

PR #881, opened 2022-02-20
fix: rename to footer.html

christophstockinger:
This file shows an example for the use of the footer component. So I have renamed the file to `footer.html`


nowseemee:
Hi @christophstockinger, thanks for the PR. This file also features the app-shell and the header so I'd prefer to keep the filename as-is.

PR #885, opened 2022-02-21
fix: text-field, textarea, date-picker, dropdown border error color

maomaoZH:


maomaoZH:
date picker error state hover and error state focus has the same border value as error state standard based on Zeplin.  I am not sure if this is intented. So I asked Joscha. Still waiting for his answer. jfyi

felix-ico:
oh ok, i think there is only one value for the error state border so should be fine

PR #887, opened 2022-02-21
fix: switch alert and warning icons

felix-ico:


acstll:
Closing until we clarify this (if needed) 🙏

PR #898, opened 2022-02-23
fix: app-logo not focusable when header is sticky

mato-a:
This fixes issue app-logo in navbar is not focusable when header is sticky


nowseemee:
Thanks for the PR!!! 🎉

Issue #882, opened 2022-02-20
Header does not behave correctly


christophstockinger:
I installed the header on my project today, but it does not behave as in Storybook. However, an analysis via Devtools shows that the styles are applied as in Storybook. 

How do I get rid of the magenta area below the menu in the sticky state?

https://user-images.githubusercontent.com/24808473/154866811-549ca8c6-c072-4f32-a555-2aec6b2e6867.mov

nowseemee:
Hi @christophstockinger, this is interesting and looks like a CSS leak to me but more information would be needed to tell - can you make a reproduction codepen, please?

christophstockinger:
Hi @nowseemee. I created a repo because I don't use codepen and have no idea how to add telekom lib.

Have a look --> https://github.com/christophstockinger/os_telekom-scale

PR #901, opened 2022-02-23
Update app-header.css

christophstockinger:


nowseemee:
Closing as https://github.com/telekom/scale/issues/882 is not a bug but a class name collision.

christophstockinger:
@nowseemee  Thx for your support. Maybe it is useful to add a namespace like `scl-` to all css classes in this project. This way there is less risk for page effects by using custom names

PR #903, opened 2022-02-23
refactor: radio button, sidebar nav, tag, footer and toggle button co…

maomaoZH:
…lor tokens


packages/components/src/components/sidebar-nav-collapsible/sidebar-nav-collapsible.css
```diff
@@ -14,7 +14,8 @@
   --border-bottom-color: var(--telekom-color-ui-additional-subtle);
   --border-left-color-third-nesting: var(--telekom-color-ui-additional-strong);
   --color-active: var(--telekom-color-text-and-icon-primary-pressed);
-  --color-primary: var(--telekom-color-text-and-icon-primary-standard);
+  --color-hover: var(--telekom-color-text-and-icon-primary-hovered);
```
felix-ico, line 17: Hi Maomao, I updated the storybook documentation with the scoped variables so we should update those in the relevant story pls have a look here for reference https://3.basecamp.com/5297099/buckets/26183347/todos/4624032394

packages/components/src/components/sidebar-nav-item/sidebar-nav-item.css
```diff
@@ -13,6 +13,7 @@
   --border-bottom-color: var(--telekom-color-ui-additional-subtle);
   --border-left-color-third-nesting: var(--telekom-color-ui-additional-strong);
   --color-active: var(--telekom-color-text-and-icon-primary-pressed);
+  --color-hover: var(--telekom-color-text-and-icon-primary-hovered);
```
felix-ico, line 16: we should update the story here as well

felix-ico (changes requested):
code changes look fine to me, we just need to update the scoped variables in the storybook documentation

felix-ico:
I didn't comment on every file but it looks like we need to update the scoped vars for every changed file except radio-button

maomaoZH:
Hi Felix I will update them and give a message when they are done! Thank you!

packages/storybook-vue/stories/3_components/toggle-group/ToggleGroup.stories.mdx
```diff
@@ -159,13 +163,83 @@ export const TemplateIcons = (args, { argTypes }) => ({
 </scale-toggle-group>
 ```
 
+### Scoped CSS variables
```
felix-ico, line 166: I think we don't need to add scoped variables if they were not already there & accessible by the user. Only update the ones that have changed from before. Pls correct me if i'm wrong @acstll

maomaoZH: I was also not completely sure about this.

Issue #917, opened 2022-03-02
Pass props to `AppLogo` on Footer


christopherGdynia:
Hello there,

the `BrandHeader` does have the options to set `href`, `title` and maybe `aria-describedby` for `AppLogo`, will the `footer `get the same features?

nowseemee:
Hello, thanks for bringing this up - yes, we will add these to the Footer as we are already preparing a new version of it.

nowseemee:
Thanks a lot for the PR @christopherGdynia - we just published the changes: `3.0.0-beta.52`. A new Storybook on B&D will follow in 24-48 hours.

Issue #916, opened 2022-03-02
Giving the header an extra attribute for accessibility


christopherGdynia:
Hello there

I can change `href`, `title` it would be nice if you can add `aria-describedby` to it.

nowseemee:
Thanks a lot for the PR @christopherGdynia - we just published the changes: `3.0.0-beta.52`. A new Storybook on B&D will follow in 24-48 hours.

PR #911, opened 2022-02-28
feat/new-logo

nowseemee:

PR #918, opened 2022-03-02
footer header props for passing to logo

christopherGdynia:
As mentioned in #917 and #916 

I have added an option, to pass `aria-describedby` and added the same functions from header to footer


christopherGdynia:
yarn format resolved in 274 file changes, should i commit it?

nowseemee:
Hi @christopherGdynia, thanks for the PR, now I understand what you've ment. Seems like your branch is behind our current main. Do you mind if I adopt your changes in https://github.com/telekom/scale/pull/911 and close this PR in favor of that?

Issue #931, opened 2022-03-09
Add additional data for pagination event


marcuskrahl:
When using the pagination component and navigating to a new page, currently only the new `startElement` is emitted in the pagination event.

However for use cases where there is a changing amount of elements per page the new `startElement` might not exactly match the desired new page.

Exaggerated Example:
Page 1: 5 elements
Page 2: 5 elements
Page 3:  20 elements (current Page)

When pressing the back button, the new `startElement` will be `0` so we reach Page 1, although we probably want to reach Page 2 when just navigating back.

It would be helpful to add a new property to the pagination event which defines what kind of navigation we performed (first, last, previous, next). That would allow consumers to correctly set pages even for a changing amount of items per page.

Old Pagination event:
```typescript
{
   startElement?: number;
}
```

New Pagination event:
```typescript
{
   direction: 'first' | 'previous' | 'next' | 'last';
   startElement?: number;
}
```

If this is an appropriate change, I will provide a PR for this change.

acstll:
Hello @marcuskrahl, thanks for getting in touch.

I also think adding this data to the event can be useful.

A PR is very welcomed 🙏

---
I would use uppercase for the values, to keep it consistent with other [similar cases in the codebase](https://github.com/telekom/scale/blob/main/packages/components/src/components/modal/modal.tsx#L31-L35). What do you think?

```
{
   direction: 'FIRST' | 'PREVIOUS' | 'NEXT' | 'LAST';
   startElement?: number;
}
```

nowseemee:
We've just published this enhancement: `3.0.0-beta.54`
Many thanks, @marcuskrahl 🚀

PR #950, opened 2022-03-18
fix(pagination): add direction to pagination events (#931)

marcuskrahl:
* fix: add a direction property to pagination events
* Fixes: #931


nowseemee:
Thanks for the PR, @marcuskrahl !!! 🎉 🎉 🎉

Issue #932, opened 2022-03-09
Scale Notifications don't have a role


christopherGdynia:
Hello there,

the notifications don't have a `role` attribute like `alert` nor a `role` `presentation` on it.

This could be an issue for accessibility

The [MUI Snackbar](https://mui.com/components/snackbars/) component and [MUI Alert](https://mui.com/components/alert/) has these roles, maybe these could help you.

```html
<!-- Alert -->
<div
  class="MuiPaper-root MuiPaper-elevation MuiPaper-rounded MuiPaper-elevation0 MuiAlert-root MuiAlert-standardError MuiAlert-standard css-xpgtxg"
  role="alert"
>
  <div class="MuiAlert-icon css-1l54tgj">
    <svg
      class="MuiSvgIcon-root MuiSvgIcon-fontSizeInherit css-1cw4hi4"
      focusable="false"
      aria-hidden="true"
      viewBox="0 0 24 24"
      data-testid="ErrorOutlineIcon"
    >
      <path
        d="M11 15h2v2h-2zm0-8h2v6h-2zm.99-5C6.47 2 2 6.48 2 12s4.47 10 9.99 10C17.52 22 22 17.52 22 12S17.52 2 11.99 2zM12 20c-4.42 0-8-3.58-8-8s3.58-8 8-8 8 3.58 8 8-3.58 8-8 8z"
      ></path>
    </svg>
  </div>
  <div class="MuiAlert-message css-1w0ym84">This is an error alert — check it out!</div>
</div>

<!-- Snackbar element -->
<div class="MuiSnackbar-root MuiSnackbar-anchorOriginBottomLeft css-qwe366" role="presentation">
  <div
    class="MuiPaper-root MuiPaper-elevation MuiPaper-elevation6 MuiSnackbarContent-root css-xl4irk"
    role="alert"
    direction="up"
    style="
      opacity: 1;
      transform: none;
      transition: opacity 225ms cubic-bezier(0.4, 0, 0.2, 1) 0ms,
        transform 150ms cubic-bezier(0.4, 0, 0.2, 1) 0ms;
    "
  >
    <div class="MuiSnackbarContent-message css-1w0ym84">Note archived</div>
    <div class="MuiSnackbarContent-action css-zykra6">
      <button
        class="MuiButton-root MuiButton-text MuiButton-textSecondary MuiButton-sizeSmall MuiButton-textSizeSmall MuiButtonBase-root css-r0mdlb"
        tabindex="0"
        type="button"
      >
[...]

PR #952, opened 2022-03-18
refactor: add role to notifications

marvinLaubenstein:
adding role and improving accessibility for notifications

(Scale Notifications don't have a role #932)


acstll:
@marvinLaubenstein is this ready for review? 👀

packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -115,15 +120,18 @@ export class NotificationBanner {
               {this.dismissible && (
                 <button
                   part="button-dismissable"
+                  type="button"
                   class="notification-banner__button-close"
                   onClick={() => this.close()}
+                  tabindex={0}
```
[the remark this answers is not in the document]
acstll: The same applies for the other 2 components…

packages/components/src/components/notification-banner/notification-banner.tsx
```diff
@@ -115,15 +120,18 @@ export class NotificationBanner {
...
+                  aria-label="close"
```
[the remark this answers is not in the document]
acstll: The same applies for the other 2 components…

acstll (changes requested):
I left a few comments, a some questions too 🤓

packages/components/src/components/notification-toast/notification-toast.css
```diff
@@ -156,22 +158,27 @@
...
 
-.notification-message__icon-close {
-  cursor: pointer;
+.notification-toast__button-close {
   position: absolute;
-  top: 8px;
-  right: 13.5px;
+  top: 10px;
+  right: 7.5px;
+  color: #191919;
```
nowseemee, line 165: can we use a token for this color?

packages/components/src/components/notification-toast/notification-toast.css
```diff
@@ -12,6 +12,7 @@
 :host {
   /* toast */
   --width: 367px;
+  --width-text-container: 271px;
```
nowseemee, line 15: this is quite a magic number - why exactly 271?

packages/components/src/components/notification-toast/notification-toast.css
nowseemee: also maybe for top and right?

packages/components/src/components/notification-toast/notification-toast.css
marvinLaubenstein: The magic number arises as a difference. So as a result of this: 
367px (total width) - 48px (icon-container) - 48px (close-button spacing equal to icon-container) = 271px 

But the question is whether the spacing on the right should always be there or only when a close button is present.

nowseemee: then why 367? 👻 you could use [calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) to make the value explicit here.

> But the question is whether the spacing on the right should always be there or only when a close button is present.

i'd say only when the button is present

packages/components/src/components/notification-toast/notification-toast.css
marvinLaubenstein: Unfortunately, we don't have tokens for that. Otherwise we would have to change the values

packages/components/src/components/notification-toast/notification-toast.css
marvinLaubenstein: 367px was the specification from design side 🫠. Added calculation instead of magic number.

Toasts must have a close button ... sorry my fault 🤦‍♂️

PR #961, opened 2022-03-29
Fix/megamenu bg

felix-ico:


packages/components/src/components/telekom/app-header/app-header.tsx
```diff
@@ -218,11 +218,8 @@ export class Header {
       readData(this.mainNavigation),
       this.activeRouteId
     );
-    const isActive = (item) =>
-      rootNode &&
-      rootNode.id === item.id &&
-      !this.visibleMegaMenu &&
-      this.visibleMegaMenu !== null;
+    const isActive = (item) => rootNode && rootNode.id === item.id;
+    console.log(rootNode, this.visibleMegaMenu);
```
nowseemee, line 222: 🎯

maomaoZH: sry i will remove it

felix-ico: this PR includes 2 fixes, one from Maomao and one from me , this code is part of @maomaoZH change :)

maomaoZH: done and pushed

nowseemee:
nice cleanup! unfortunately the rounded separator is removed with these changes. is it intentional? 
image

felix-ico:
> nice cleanup! unfortunately the rounded separator is removed with these changes. is it intentional? image

ah thanks for checking i didn't notice that in light mode, fixing it

nowseemee:
Dark mode was like this: 
image

felix-ico:
ok I added the separator and those "ghost" elements to make the border radius - i had removed too much code previously!

packages/components/src/components/telekom/app-header/app-header.tsx
```diff
@@ -218,11 +218,7 @@ export class Header {
       readData(this.mainNavigation),
       this.activeRouteId
     );
-    const isActive = (item) =>
-      rootNode &&
-      rootNode.id === item.id &&
-      !this.visibleMegaMenu &&
```
nowseemee, line 224: are you sure this was dead code? you can never know. can we keep it, just to be sure? the header has so many use-cases.

felix-ico: @maomaoZH i think this was part of your bugfix, what do you think?

maomaoZH: So the designer wants to keep the selected nav status even when user navigates to other tab. If we keep the code, every time when user navigates to other tab which has mega-menu, the selected status will lose.

maomaoZH: give me one sec please. dont merge yet

maomaoZH: https://user-images.githubusercontent.com/23239723/160800784-c36f829e-5d43-4627-b7e7-901d73e968cc.mov

have you noticed the menu is jumping.

felix-ico: I pushed a fix for this, thanks for checking!

PR #964, opened 2022-03-30
fix: update react-component-lib

felix-ico:


felix-ico:
Thanks just pushed a prettier fix. 

I tested `yarn build:components`, is that what you mean?

acstll:
> I tested yarn build:components, is that what you mean?

yas! 👍

PR #965, opened 2022-03-31
fix: add hover color to flyout menu with focus status

maomaoZH:


maomaoZH:
For the fix : https://3.basecamp.com/5297099/buckets/26183347/todos/4778741695#__recording_4791570056

nowseemee:
Thanks for the PR, @maomaoZH !!! 🎉 🎉 🎉

PR #988, opened 2022-04-11
feat: add min max attribute to text-field

maomaoZH:


packages/components/src/components/text-field/text-field.tsx
```diff
@@ -45,11 +45,15 @@ export class TextField {
...
+    | 'month' // example yyyy-mm
+    | 'week' // example yyyy-W##
+    | 'time' // example hh:mm
+    | 'datetime-local' // example yyyy-mm-ddThh:mm
     | 'url' = 'text';
   /** (optional) Input name */
   @Prop() name?: string = '';
   /** (optional) Input label */
-  @Prop() label: string = '';
+  @Prop() label?: string = '';
```
acstll, line 56: I think the `label` should be mandatory because of accessibility 🤔

acstll (commented):
This is great! thank you @maomaoZH 

I only left a question regarding the label…

packages/components/src/components/text-field/text-field.tsx
maomaoZH: hmm, that means if label is not set, then we set default as empty?

maomaoZH: i updated it. I remove `optional`

Issue #984, opened 2022-04-08
Modal dialog accessibility issue with scrolling


mato-a:
Hi all,

I'd like to report accessibility bug in modal dialog related to scrolling behavior.

When modal dialog is opened and you scroll on it with mouse to bottom and keep scrolling, screen under modal window will scroll down. When you scroll with mouse to top of the window and keep scrolling, screen under modal window will scroll up. Same applies when you use up and down arrow keys.

It is also possible to scroll background content when modal dialog doesn't have any scrollbar of its own - both with mouse and up and down arrow keys. 

According the specs the windows under a modal dialog are inert: https://www.w3.org/TR/wai-aria-practices/#dialog_modal

Sandbox: https://codesandbox.io/s/vanilla-js-scale-7vt842

mato-a:
Hi @felix-ico, the issue can be reproduced also in the lastest version: `3.0.0-beta.103`

mato-a:
Hi @felix-ico, 
Thanks, it works after adding the attribute `allow-injecting-style-to-body`.
Yeah, I think that the default should be set to `true`.

mato-a:
Hi @felix-ico 

There is another issue related to this. The css rule `overflow-y: auto` is set to the body of the modal, which makes it scrollable. However there is an issue how to make this scrollable with the keyboard. Firefox automatically makes the scrollable div focusable by keyboard navigation, Chrome does not. 

There is an official proposal provided by w3: https://www.w3.org/WAI/standards-guidelines/act/rules/0ssw9k/proposed/

verdody:
Hi. The issue of the last comment from @mato-a is still persisting. Do you have a solution for this?

PR #1003, opened 2022-04-21
fix: clicking esc key to close all opened modal window

maomaoZH:


packages/components/src/components/modal/modal.tsx
```diff
@@ -190,8 +191,16 @@ export class Modal {
...
       this.open();
+      if (this.allowInjectingStyleToBody) {
+        // The following style will disable body from scrolling when modal is open
+        document.body.style.overflow = 'hidden';
+      }
     } else {
       this.close();
+      if (this.allowInjectingStyleToBody) {
+        // remove injected css style from body
+        document.body.style.overflow = null;
```
acstll, line 202: …and set it back to its previous value, or `null` if there was nothing.

acstll (commented):
Thank you @maomaoZH 🙌

packages/components/src/components/modal/modal.tsx
```diff
@@ -155,6 +155,7 @@ export class Modal {
       this.resizeObserver.observe(this.modalBody as Element);
     }
     this.setHasScroll();
+    this.hostElement.addEventListener('keydown', this.handleKeypress);
```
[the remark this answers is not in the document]
maomaoZH: sounds good!  I will register the listener when it opens and remove it when it  closes.

packages/components/src/components/modal/modal.tsx
maomaoZH: yeah, sure! I will adjust them.

packages/components/src/components/modal/modal.tsx
```diff
@@ -192,14 +193,15 @@ export class Modal {
     if (newValue === true) {
       this.open();
       if (this.allowInjectingStyleToBody) {
+        this.bodyOverflowValue = document.body.style.overflow;
```
maomaoZH, line 196: I store the value in a state. If there is no overflow value is set, it will store at empty string.

maomaoZH:
@acstll  Hi, Arturo, I did some update changes.

nowseemee:
Is the PR https://github.com/telekom/scale/pull/1013 addressing something similar? @felix-ico

felix-ico:
@nowseemee that PR is for https://github.com/telekom/scale/issues/984 similar but not the same

nowseemee:
> @nowseemee that PR is for #984 similar but not the same (only prevent the body from scrolling when scrolliing on the modal)

Can you please explain me the difference? You set `this.hostElement.closest('body').style.overflow` and Maomao sets `document.body.style.overflow`.

felix-ico:
@nowseemee I see, perhaps this PR is handling two issues... I guess mine can be closed as this seems more solid

felix-ico:
sorry for commenting on a closed PR but this is related https://github.com/telekom/scale/issues/984, what do you think of setting the prop to true by default?

PR #1006, opened 2022-04-21
fix: remove transform to maintain fixed positioning in child elements

felix-ico:


packages/components/src/utils/animate.ts
```diff
@@ -61,13 +61,13 @@ export const KEYFRAMES = {
       offset: 0,
       ...keyframeDefaults,
       opacity: 0,
-      transform: 'translateY(-3rem)',
+      top: '-3rem',
```
acstll, line 64: I would probably add a comment here explaning why were not using `transform` with a link to that piece of mdn regarding `fixed`? — I can imagine somebody legitimately changing this back in the name of performance 🤔

acstll (approved):

packages/components/src/utils/animate.ts
felix-ico: makes sense, i added a comment there

Issue #989, opened 2022-04-11
Modal dialog accessibility issue with Escape key


matandre1:
Hello guys,

I found accessibility problem when closing modal dialog with _'Esc'_ key.

**Current behavior:** I have opened modal window which open another modal window. Now I press Esc key and all modal windows will be closed.

**Expected behavior:** When modal window open another modal window Esc key close only latest modal dialog and all other have to be opened.

**Sand box example:** https://codesandbox.io/s/jdbbsl?file=/index.html

Thank you.

Best regards,
Matus

PR #1013, opened 2022-04-26
fix: hide body overflow when modal is opened

felix-ico:


felix-ico:
closing this in favor of https://github.com/telekom/scale/issues/989

PR #1015, opened 2022-04-26
Fix/header html

marvinLaubenstein:
Content of the pull request: 

- ✅ fix of issue "Examples from Storybook are incorrect and not accessabilty conform #879" 

( Accessibility was checked with Jaws 2019 and no negative change could be detected. Please cross check again before accepting the pull request )


nowseemee (approved):
awesome! thanks a lot Marvin! 🙌🏻

PR #1019, opened 2022-04-28
fix: remove padding, add font-weight to text inpout label

felix-ico:


nowseemee:
many thanks for the PR! could you please update the visual snapshots?

felix-ico:
updated!

Issue #1009, opened 2022-04-22
scale-app-header covers scale-menu-flyout-list when part of scale-menu-flyout-list is crossed with scale-app-header


Tireile:
Hello!

It looks like there is wrong displaying of scale-menu-flyout-list in case when it is crossed with scale-app-header. list should be rendered above header. It happens when scale-menu-flyout-list has no enough place to render under trigger button.

Sandbox: https://codesandbox.io/s/confident-galois-97usmc?file=/src/App.vue

![scaleBug](https://user-images.githubusercontent.com/35307181/164725421-fced0af0-b330-4efa-bf29-86904a5c25f4.gif)

felix-ico:
Hi @Tireile, thank you for reporting this issue. I am having a look at this

nowseemee:
release coming soon

felix-ico:
this is now fixed in 3.0.0-beta.103

Tireile:
@felix-ico, bug is not fixed
![image](https://user-images.githubusercontent.com/35307181/166930584-592cbb0c-5482-41d5-b99a-d8dbc604aaa7.png)

felix-ico:
Hi @Tireile could you please have another look? I forked your sandbox and updated scale-components to 3.0.0-beta.103
https://codesandbox.io/s/hopeful-dewdney-5r1638?file=/package.json

<img width="1371" alt="Screenshot 2022-05-05 at 15 50 02" src="https://user-images.githubusercontent.com/97440128/166937701-a7b911ea-2994-4d56-bf55-d3dca99be953.png">
<img width="1087" alt="Screenshot 2022-05-05 at 15 51 52" src="https://user-images.githubusercontent.com/97440128/166938065-f6cc30e9-1f47-4b6d-9358-9995ca20b288.png">

Tireile:
Hello, @felix-ico. I found out what was a problem in our side. We use scale-flyout-list inside component with position: sticky, so that's why list was covered by header. Thanks for the feedback

PR #1022, opened 2022-04-29
fix: increase flyout zindex

felix-ico:
fixes https://github.com/telekom/scale/issues/1009


packages/components/src/components/menu-flyout-list/menu-flyout-list.css
```diff
@@ -17,7 +17,7 @@
 :host {
   box-sizing: content-box;
   position: fixed;
-  z-index: var(--scl-z-index-20);
+  z-index: 100;
```
acstll, line 20: I would add a comment here, something like: /** Above the brand header (a solid system for z-index values would be nicer) */

Issue #1036, opened 2022-05-06
ScaleTooltip causing layout issues


Darkavid:
Hello Scale Team,

I'm trying to use the `ScaleTooltip` component. I know it's still in Beta but perhaps you can help me out anyway.

It provides a fancy appearing tooltip box on hover, which is cool, however it messes up the available width space and styling of its children.

Normal `ScaleTextField` (code):
```
<>
        <ScaleTextField name="test" label="Test" value="Text field value" />
</> 
```

Normal `ScaleTextField` (result):
![image](https://user-images.githubusercontent.com/28023185/167103059-953705cf-5d23-4cb5-bc64-49d65af5c05e.png)

Tooltipped `ScaleTextField` (code):
```
<>
      <ScaleTooltip content="Tooltip content">
          <ScaleTextField name="test" label="Test" value="Text field value" />
      </ScaleTooltip>
<>
```

Tooltipped `ScaleTextField` (result):
![image](https://user-images.githubusercontent.com/28023185/167103502-1af067ee-ce24-4d5d-9012-04d9be62426b.png)

I'm not sure if it is intentional but I do not find it to be intuitive from a developer POV.
Maybe you understand that my intention was to put a tooltip onto the children without changing its width.

It is even worse with `ScaleDropdown`.

Normal `ScaleDropdown` (code):
```
<>
        <ScaleDropdown name="test" label="Test" />
</> 
```

Normal `ScaleDropdown` (result):
![image](https://user-images.githubusercontent.com/28023185/167103760-cbaa86f9-e280-42df-a392-52a0faad0270.png)

Tooltipped `ScaleDropdown` (code):
```
<>
      <ScaleTooltip content="Tooltip content">
           <ScaleDropdown name="test" label="Test" />
      </ScaleTooltip>
</>
```

Tooltipped `ScaleDropdown` (result):
![image](https://user-images.githubusercontent.com/28023185/167103892-4732016f-cd55-448e-95da-08c796d5ab84.png)

Could you please tell me how can I use ScaleTooltip **without messing with the layout**?

felix-ico:
Hi @Darkavid thank you for reporting this. I could confirm the bug and am looking for a cause and fix

Darkavid:
Hi Team,

Is there any update on this issue?

Thank you!

Tireile:
@felix-ico Hello, do you have any updates?

nowseemee:
[...]

PR #1039, opened 2022-05-06
fix: avoid resizing child elements like dropdowns or input fields

felix-ico:
fix for https://github.com/telekom/scale/issues/1036


felix-ico:
had not tested my fix with the button, closing for now

PR #1048, opened 2022-05-10
docs: allow injecting styles from storybook

felix-ico:


felix-ico:
as discussed, perhaps we should remove the style injection from storybook for all components, instead of adding it to all components. Closing this stale PR

Issue #1016, opened 2022-04-26
<scale-link> is not compatible with <scale-breadcrumb> 


adamfeher64:
This works:
``` html
<scale-breadcrumb>
  <a href="#1"> First link </a>
  <a href="#2"> Second link </a>
  <a href="#3"> Third link </a>
</scale-breadcrumb>
```
Unexpectedly, this doesn't work:
``` html
<scale-breadcrumb>
  <scale-link href="#1"> First link </scale-link>
  <scale-link href="#2"> Second link </scale-link>
  <scale-link href="#3"> Third link </scale-link>
</scale-breadcrumb>
```

marvinLaubenstein:
Hi @adamfeher64, I could now take a look at it. 

Unfortunately, I could not reproduce the error. 

![Bildschirmfoto 2022-04-27 um 13 39 49](https://user-images.githubusercontent.com/82942834/165511303-c6a0550a-31cb-4793-b419-1e0928ba7822.png)
<img width="284" alt="Bildschirmfoto 2022-04-27 um 13 39 04" src="https://user-images.githubusercontent.com/82942834/165511308-724464b3-6a68-4699-aa28-7a2664497698.png">

Could you maybe give us some more information. For example browser in which the error occurs and your scale version, maybe also your wrapping code, in which the **scale-breadcrump** was integrated. Thank you :)

Regards,
Marvin

adamfeher64:
Hi @marvinLaubenstein, 
I'm using Scale version 3.0.0-beta.102 installed via NPM, integrated into Angular project with version 13.2.4. I'm trying to implement `<scale-breadcrump>` tag into `menu-main` slot of `<scale-app-header>` tag just like this:
``` html
<scale-app-header slot="header">
  <div slot="menu-main">
    <scale-breadcrumb>
      <scale-link omit-underline [routerLink]="['/en', 'first']">
        <span> Link #1 </span>
      </scale-link>
      <scale-link omit-underline [routerLink]="['/en', 'second']">
        <span> Link #2 </span>
      </scale-link>
    </scale-breadcrumb>
  </div>
</scale-app-header>
```
The code above is saved as Angular component with `<app-header>` tag selector and injected into `<scale-app-shell>` tag, just like this:
``` html
<scale-app-shell>
  <app-header slot="header"> </app-header>
  <div class="content">
    <router-outlet></router-outlet>
  </div>
  <app-footer slot="footer"></app-footer>
</scale-app-shell>
```
[...]

PR #1089, opened 2022-06-01
feat: adding possibility to use a tags with scale-link inside breadcrumbs

marvinLaubenstein:
fix: <scale-link> is not compatible with <scale-breadcrumb> #1016 

@acstll (we talked about this before at Basecamp)

### DIFF
--- a/packages/components/src/components/breadcrumb/breadcrumb.tsx
+++ b/packages/components/src/components/breadcrumb/breadcrumb.tsx
@@ -91,7 +91,7 @@ export class Breadcrumb {
                 );
               return (
                 <li class="breadcrumb__list-item" part="list-item">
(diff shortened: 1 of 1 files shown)


acstll (approved):
@marvinLaubenstein thanks for this!

Just to make sure I get it correctly, this is "partial" fix, because using `scale-link` inside the breadcrumbs components will render an actual link (not a `span`), but the styles won't match the ones from `scale-link`, correct?

marvinLaubenstein:
Thanks for reviewing @acstll .

Yes exactly, the solution is first of all a nice to have, so that users can use `scale-links` in the `scale-breadcrumb` and not automatically a `span` arise although you wanted a link. 

Final solution, which unfortunately has not yet been clarified, would of course be to remove the `a` tag and replace it with our `scale-link `

PR #1090, opened 2022-06-02
fix: fixing possible default z-index problems

marvinLaubenstein:
⚠️ Please check if this is wanted like this before merging. 

current status:
Bildschirmfoto 2022-06-02 um 21 03 25
with changed z-index by variable:
Bildschirmfoto 2022-06-02 um 22 00 35

I noticed a problem with the fixed z-indexes (compare screenshot). I think it can definitely be useful for the user of the component if he can set the z-index variably. 

In the variables I have set the old values as default these change nothing and could also be 0. But someone has certainly thought something with the values.


packages/components/src/components/date-picker/date-picker.css
```diff
@@ -100,6 +102,7 @@ duet-date-picker .duet-date__toggle:active scale-icon-content-calendar {
 }
 
 .scale-date-picker {
+  --z-index-label: 10;
```
nowseemee, line 105: should we add this to the top of the file, or is there a reason you've inserted a new variable into this position? thanks :)

acstll: @marvinLaubenstein I agree with Daniel this line should be moved up, to place it inside `:host`, around line 34. Other than that, this looks fine!

acstll (changes requested):
Left a comment, I think the `--z-index-label` declaration should be placed in the :host. Or is there a reason not to do it?

nowseemee:
@acstll do you approve?

acstll:
Will also take a look @nowseemee — sorry I missed the mention (I need to fix my email/notification/etc.) 😅

acstll (approved):

acstll:
done… @marvinLaubenstein LGTM

PR #1117, opened 2022-06-30
Refactor/tab nav

marvinLaubenstein:
Component changes requested from the design side:

Bildschirmfoto 2022-06-30 um 08 48 09

@acstll


acstll (changes requested):
Thank you @marvinLaubenstein, here's what I found, I hope it helps.

Visual:
- tab-header: spacing (padding) is too little, I think height should be 48px (screenshot)
- tab-header: right margin is not needed (bottom border is continuous)
- tab-header: border color is too light?
- tab-nav?: bottom border should also reach the end of the component on the right

Implementation:
- regarding the props, I would do the opposite and keep the deprecated `small` prop, and deprecate the `size` prop — we leave it but won't do anything except for the console.log message (we don't want to introduce any breaking changes until v4)
- I think the CSS variables are not properly documented in Storybook

Captura de pantalla 2022-07-01 a les 10 10 53

packages/components/src/components/tab-header/tab-header.tsx
```diff
@@ -33,10 +33,8 @@ export class TabHeader {
   /** True for a disabled Tabnavigation */
   @Prop() disabled?: boolean = false;
   /** True for smaller height and font size */
-  // DEPRECATED - size should replace small
+  /** @deprecated - css overwrites should replace small */
   @Prop() small?: boolean = false;
-  /** (optional) size  */
-  @Prop() size: 'small' | 'large' = 'large';
```
acstll, line 39: This we want to keep (so it's not a "breaking" change) 🙂

We add the `statusNote` the same way as with the `small` prop (and the "DEPRECATED" comment).

So if somebody uses either `size` or `small` they will get the warning in the console. But the prop has actually no effect on the component.

acstll (changes requested):
Sorry if this was not clear enough in my previous round of feedback 🙏

marvinLaubenstein:
⚠️ Please do not merge before all other size changes in the components are available

packages/components/src/components/tab-header/tab-header.tsx
marvinLaubenstein: Had already wondered extremely about this. Thanks for the clarification 🙂

Issue #1096, opened 2022-06-10
Menuflyoutlist  Cannot read properties of undefined (reading 'forEach')


christopherGdynia:
**Scale Version**
beta 102

**Framework and version**
react 17
scale-*-react beta 102

**Current Behavior**
A clear description of what the bug is and how it manifests.

**Expected Behavior**

**Code Reproduction**
A minimal reproduction case of the bug, you may fork one of the available sandboxes as a base: 

**Additional context**

```
// stacktrace  
next-dev.js?b90a:32 TypeError: Cannot read properties of undefined (reading 'forEach')
    at MenuFlyout.closeAll (app-mega-menu_11.entry.js?6b7b:390:1)
    at MenuFlyout.handleWindowScroll (app-mega-menu_11.entry.js?6b7b:354:1)
    at eval (index-f2e5998d.js?737d:51:1) undefined
```

```js
// this.lists is undefined

closeAll() {
    this.lists.forEach(async (list) => {
      await list.close(); // Wait for `scale-close` event to fire
      list.active = false; // Make sure focus control is right while reopening
    });
  }

```

![image](https://user-images.githubusercontent.com/84845536/173045062-4deb9201-5065-4be2-8943-af7950b468c5.png)
![image](https://user-images.githubusercontent.com/84845536/173045127-d102fef7-ac16-46f8-8523-7d8da4368a83.png)

PR #1118, opened 2022-06-30
fix(flyout): avoid type errors list

nowseemee:
it's most likely a `this` problem but could not reproduce.
https://github.com/telekom/scale/issues/1096

### DIFF
--- a/packages/components/src/components/menu-flyout/menu-flyout.tsx
+++ b/packages/components/src/components/menu-flyout/menu-flyout.tsx
@@ -126,7 +126,7 @@ export class MenuFlyout {
   }
 
   closeAll() {
-    this.lists.forEach(async (list) => {
+    (this.lists || new Set()).forEach(async (list) => {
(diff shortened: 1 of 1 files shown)


nowseemee:
@acstll could you have a look, please?

packages/components/src/components/menu-flyout/menu-flyout.tsx
```diff
@@ -126,7 +126,7 @@ export class MenuFlyout {
   }
 
   closeAll() {
```
acstll, line 128: I would additionally make this an "arrow" function, just like the `toggle` below…

```diff
-closeAll() {
+closeAll = () => {
```

if it's really a `this` problem, this would probably fix it (pun intended haha)

acstll (commented):
Posted one suggestion, what do you think?

acstll (approved):
👌

PR #1120, opened 2022-07-01
feat(button): forms will submit with Enter key

### LINKED ISSUE #859: Form submit not triggered on enter
In a form that contains multiple scale-text-field ans a scale-button of type submit, hitting enter in any of the inputs does not trigger the submit of the form as it should be.

acstll:
Fixes #859.

- Not a "React wrapper" thing
- Same problem with "generic" web component
- Forms with a single text input will be submitted with the Enter key, even without a button: https://stackoverflow.com/a/35235768

---

- [x] fix
- [x] unit or e2e test
- [ ] find a better word instead of "fallback"?

### DIFF
--- a/packages/components/src/components/button/button.e2e.ts
+++ b/packages/components/src/components/button/button.e2e.ts
@@ -38,4 +38,20 @@ describe('scale-button', () => {
     const element = await page.find('scale-icon-action-search');
     expect(element.getAttribute('size')).toBe('16');
   });
+
+  it('should allow submitting forms with the Enter key', async () => {
+    const page = await newE2EPage();
+    await page.setContent(`
+      <form>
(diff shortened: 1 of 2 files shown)


nowseemee:
My five cents: I would add the [form behavior findings](https://github.com/telekom/scale/issues/859#issuecomment-1173599883) to our docs and point people to it instead of extending the button component with this.

nowseemee (approved):
And this code is not wrong either - not easy to decide what's the correct thing to do.

acstll:
@nowseemee I think I'm OK with merging this. Polished it a tiny bit. Any final thoughts? 🤠

nowseemee:
let's go 🫣

PR #1126, opened 2022-07-07
feat(menu-flyout): close menu with click on trigger

### LINKED ISSUE #922: a11y: Close Flyout Menu on second click of trigger
Currently, subsequent clicks on the trigger button of a Flyout Menu does not do anything.

Since the trigger is not a hover event but a toggle button, it should act like a toggle for good accessibility & ergonomics. Meaning that if the Flyout Menu is currently open, a subsequent click should close the Flyout Menu.

This is less relevant for keyboard navigation, because after opening the Flyout Menu, the focus is trapped in the list. Closing the Flyout Menu then is possible via ESC key - which is ok.

Reproduction possible with the standard docs example:
https://www.brand-design.telekom.com/scale/?path=/docs/components-flyout-menu--standard

acstll:
Fixes #922, ~~and also 1122!~~

@marvinLaubenstein please 👀

### DIFF
--- a/packages/components/src/components/menu-flyout/menu-flyout.tsx
+++ b/packages/components/src/components/menu-flyout/menu-flyout.tsx
@@ -47,6 +47,25 @@ export class MenuFlyout {
 
   private trigger: HTMLElement;
   private lists: Set<HTMLScaleMenuFlyoutListElement> = new Set();
+  // Keep track of the current active/open list
+  private activeList: HTMLScaleMenuFlyoutListElement;
+
+  @Listen('scale-open')
+  async handleScaleOpen({ detail }) {
+    // Close the previous active list if
+    // - it's not the root and
+    // - it's not the one being opened
+    // (useful only with "click" interactions)
+    if (
+      this.activeList &&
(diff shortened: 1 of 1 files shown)


marvinLaubenstein (dismissed):
Hi @acstll. It seems to work for a sublist. However, if they are further nested, it does not work anymore .

https://user-images.githubusercontent.com/82942834/177854572-c1148a86-2d4d-45b7-a18d-722e51475ed0.mov


I have also tried my luck (https://github.com/telekom/scale/tree/fix/sub-list-problems). Everything worked like it should but I have a double click bug that I can not get rid of. 

But I'm already super excited about your solution. (Mine is rather a little bit tinkered together) 🤓

acstll:
Oh, thanks a lot @marvinLaubenstein 🙌

Then let's focus this PR only on #922. I'll revert 100db74a119186f9e15eaad04ebbd92795dfbda8 now.

acstll:
@marvinLaubenstein this time only focusing on #922 🙏

marvinLaubenstein (approved):

PR #1127, opened 2022-07-07
feat(modal): add --min-height-window var

### LINKED ISSUE #731: Scale Modal --min-height-window
In the scoped variables there is a --max-height-window, but why isn't there a possibility to set the min height of a modal?
In some cases this could be useful.

acstll:
Closes #731 (it didn't hurt)


acstll:
@nowseemee 👀

nowseemee (approved):
👍🏻

PR #1129, opened 2022-07-08
fix: use next release version for working nuxt3 app

felix-ico:
The boilerplate at the moment only works by yarn linking the scale-components using the latest stencil version. Updating its scale-components version to the next release version so that the example will actually work "out of the box"


acstll (changes requested):
👀

acstll (approved):
👌

Issue #805, opened 2022-01-20
React: Funky reactivity / component value binding


defaude:
Both inputs in [App.tsx](https://github.com/defaude/scale-reactivity/blob/main/src/App.tsx) are supposed to "reject" all "a" characters (i.e. simply remove it immediately when one is detected. This is achieved using an effect that fires whenever the text value is updated. The normal HTML input works just fine: Trying to type an "a" in there just won't do.

However: When using the Scale TextField, the effect is triggering correctly (state value gets updated properly) but the Scale TextField still shows the "a" character(s) at the end. It seems that the value change from the outside is not properly reflected into the component's internal state?

Going one step further: The second pair of inputs is bound so that only their `value` is bound but no change handler is registered. (Yeah, we could discuss the sense or meaning of this, of course, but let's just stick with the tech for a second 😅). The regular input element will be effectively "read only" since the underlying React value doesn't change. The Scale TextView component, however, allows editing - at least "optical" since the state changes are not sent to the React app, of course.

christopherGdynia:
Hello @defaude,

can you show this in a [codesandbox](https://codesandbox.io) ?

christopherGdynia:
Nevermind, I found your repos and created a Sandbox with your code [here](https://codesandbox.io/s/stoic-noyce-0f4v9?file=/src/index.tsx:218-222)

christopherGdynia:
Ok, after some playing around with it, I am still clueless, the ```ScaleTextField``` detects changes on the value prop and handles it.

I have tried to work with the ```onKeyDown``` eventlistener to filter the input with ```event.key``` 
The performance was a bit better, but still not good.
And the ```ScaleTextField``` onChange eventlistener triggered everytime

defaude:
Yeah I've played around with different event types etc. **But** the general issue is that the Scale components seem to detach their internal state from the values that are given from the outside world.

After typing an "a" into the `ScaleTextField`, the change handler properly fires. Which in turn updates the `text` state, which triggers the effect to run, which re-updates the `text` state once again (with the string cleaned from all "a" instances). However, the `ScaleTextField` just doesn't pick up that last change.

acstll:
Sorry for the super-late reply. @defaude `ScaleTextField` cannot be "controlled" like a regular `input`, this I'm 99% certain. So the 2 things I would like to figure out would be: (1) is the wrapper to blame for this? and (2) can we make this work without the component being "controlled"?

[...]

PR #1140, opened 2022-07-15
fix: add controlled prop

felix-ico:
So using the same approach as in the dropdown fixes this [react/reactivity issue](https://github.com/telekom/scale/issues/805), I guess this would need to be added to some other components as well to avoid the same issue in the future (at least in text-area).


packages/components/src/components/text-field/text-field.tsx
```diff
@@ -156,6 +168,11 @@ export class TextField {
 
   handleInput = (event: Event) => {
     const target = event.target as HTMLInputElement | null;
+    if (this.controlled) {
+      emitEvent(this, 'scaleChange', { value: target.value });
```
acstll, line 172: Is firing `scaleChange` here really needed? it's being fired again on line 178. Also, doesn't the re-render triggered by `forceUpdate` fire another change event?

Maybe it'd make sense to add a test to check whether the change event is not being fired too much?

acstll (commented):
Also, should we also document this new `controlled` prop in Storybook? I think yes, but with a notice saying it's "experimental"… maybe?

packages/components/src/components/text-field/text-field.tsx
felix-ico: yup that was a mistake, removed and added a test

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -96,6 +96,8 @@ export class TextField {
   /** (optional) Injected CSS styles */
   @Prop() styles?: string;
 
+  /** (optional) Makes type `input` behave as a controlled component in React */
+  @Prop() controlled?: boolean = false;
```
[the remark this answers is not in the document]
felix-ico: so i replaced (optional) with (experimental) but I'm not sure if that is what you meant or to change the prop name to something like `experimental_controlled` ? Wasn't sure because in other components like the dropdown there is also a `controlled` prop but without any experimental prefix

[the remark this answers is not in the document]
acstll: yes, haha, I was thinking about `experimental_controlled` — the old Dropdown is going to be deprecated.

Would it work with the snake_case? 🤔

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -95,7 +95,8 @@ export class TextField {
 
   /** (optional) Injected CSS styles */
   @Prop() styles?: string;
-
+  /** (experimental)) Makes type `input` behave as a controlled component in React */
```
acstll, line 98: there's an extra ) in `(experimental))` 🤓

packages/components/src/components/text-field/text-field.tsx
acstll: amazing, thank you!

packages/components/src/components/text-field/text-field.tsx
felix-ico: > Would it work with the snake_case? 🤔

seems like it doesn't work, the linter complains. experimentalControlled/experimental-controlled works

acstll: I'm not in love with `experimental-controlled` but I think I prefer it over just `controlled` (only because it's a React-only thing)

acstll:
I got confused with the many "conversations" I opened in the comments, haha, sorry 🙏

felix-ico:
@acstll all CI jobs are passing now, snake_case doesn't work so i changed it to experimental-controlled. 

I also added the prop to storybook, but there is no real documentation on why/how someone would use it. Should that be included in storybook as well? or maybe just in the readme?

acstll (approved):
🙈

PR #1155, opened 2022-07-28
Fix/scale spacing

felix-ico:
replacing all leftover --scl variables, not only spacing variables


felix-ico:
in data-grid.css i replaced the headline typography styles with the new `--telekom-text-style` tokens as semantically they seemed to fit. However, the values are different, and, as discussed elsewhere, i guess those are not actually meant to be used.

acstll:
> in data-grid.css i replaced the headline typography styles with the new `--telekom-text-style` tokens as semantically they seemed to fit. However, the values are different, and, as discussed elsewhere, i guess those are not actually meant to be used.

the values are different because there was an visual update in the styles themselves — this (data-grid.css) is the only place in codebase I could think of we could use this `--telekom-text-style` tokens, so this is perfect 👍

acstll (approved):
Thanks a 💯 for this!

PR #1160, opened 2022-08-03
docs: add vue3 vite boilerplate

felix-ico:


examples/vue3-vite-boilerplate/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@telekom/scale-components": "^3.0.0-beta.110",
```
acstll, line 12: We use the neutral version (`@telekom/scale-components-neutral`) in all examples so people is on the safe side legally when just "trying it out"

https://github.com/telekom/scale/blob/main/examples/vanilla/index.html#L13

examples/vue3-vite-boilerplate/src/App.vue
```diff
@@ -0,0 +1,28 @@
...
+  padding: 1.5em;
+  will-change: filter;
+}
+.logo:hover {
+  filter: drop-shadow(0 0 2em #646cffaa);
+}
+.logo.vue:hover {
+  filter: drop-shadow(0 0 2em #42b883aa);
+}
+@import "@telekom/scale-components/dist/scale-components/scale-components.css";
```
acstll, line 26: Is this really needed here if we're already importing it on `main.js`?

examples/vue3-vite-boilerplate/src/style.css
```diff
@@ -0,0 +1,90 @@
...
+  min-width: 320px;
+  min-height: 100vh;
+}
+
+h1 {
+  font-size: 3.2em;
+  line-height: 1.1;
+}
+
+button {
```
acstll, line 49: Maybe some of these CSS can be removed if not needed (it'd be nice there was only what's needed for the page to look decent but nothing else)…

examples/vue3-vite-boilerplate/src/App.vue
```diff
@@ -0,0 +1,28 @@
...
+    <a href="https://vitejs.dev" target="_blank">
+      <img src="/vite.svg" class="logo" alt="Vite logo" />
+    </a>
+    <a href="https://vuejs.org/" target="_blank">
+      <img src="./assets/vue.svg" class="logo vue" alt="Vue logo" />
+    </a>
+  </div>
+  <scale-button>Hello Scale</scale-button>
+  <span> hello </span>
+</template>
```
acstll, line 12: If we could make this look like this, it'd be great https://github.com/telekom/scale/blob/main/examples/angular-boilerplate/src/app/app.component.html

acstll (commented):
This looks good to merge already, but I think it can be a bit better 🤓

examples/vue3-vite-boilerplate/package.json
felix-ico: 👍 updated

examples/vue3-vite-boilerplate/src/App.vue
felix-ico: Updated this, I also noticed that the vanilla boilerplate wasn't rendering the scale button correctly anymore so pushed a fix for that too

examples/vue3-vite-boilerplate/src/style.css
felix-ico: removed unnecessary code

acstll:
@felix-ico LGTM 👌

PR #1164, opened 2022-08-05
refactor(tooltip): update color tokens, arrow size

acstll:
According to the latest visual spec…

- [x] update color tokens
- [x] adjust arrow size and placement

I used pixels for the arrow size, which might be against our convention but for this particular case (considering the offset config) I think it makes sense. It doesn't hurt accessibility.

(Another PR will follow after this is merged.)

@nowseemee may you please… 👀


acstll:
⚠️ Gonna give it a try with the visual tests that surely need to be updated — I might need assistance from a kind soul.

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -67,6 +67,7 @@
 
 [part='arrow'] {
   position: absolute;
+  z-index: -1; /* always below the text */
```
nowseemee, line 70: so this might cause the arrow to disappear under another element, right? The tooltip itself uses `--scl-z-index-70` - shall we just make it `70-1` for the arrow?

acstll: nice catch, I'm going to test this… I think that -1 might be relative to 70 (within the "scope" of the tooltip element) 🤔

acstll: works as expected, can stay `-1` 👍 — thanks again for asking

nowseemee (approved):

PR #1165, opened 2022-08-08
fix(tooltip): a few issues

acstll:
- [x] use trigger element (child inside slot) for events, and not the host itself, to avoid margins triggering the tooltip
- [x] warn if used without a trigger element (just text, wrong usage)
- [x] add beta notice
- [x] fix jump when visible early (the element was visible before it was positioned)
- [x] use `pointer-events: none` when hidden, to avoid interfering
- [x] rename `open` to `opened` for consistency with other components (Modal et al) — ⚠️ breaking

@nowseemee if you're so kind 👀 — @marvinLaubenstein please also take a look if you have time 🙏


marvinLaubenstein (commented):
> Really nice 🙂, but shouldn't the prop renaming from `open` to `opened` also be in the storybook ?

nice catch 🙌 — I completely forgot

nowseemee (approved):
thanks for adding these! 🙌🏻

acstll:
I need yer help @nowseemee 🤠

image

nowseemee:
> I need yer help @nowseemee 🤠
> 
> image

on it 🤠

PR #1170, opened 2022-08-16
refactor(tooltip): a11y 💙

acstll:
https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/

- [x] right tokens
- [x] keyboard interaction
- [x] ARIA roles, states and props
- [ ] stories make sense (nothing overly complicated, nothing superficial)
- [ ] has usage text


acstll:
Merging the fix without removing the `beta` label yet. Another PR will follow. Should fix https://github.com/telekom/scale/issues/1225.

PR #1179, opened 2022-08-23
brand header profile icon differ from default and selected version

niclas18:
The profile-icon in the brand header change at OnClick. The switch use to icons named `default.svg` and `selected.svg` . The selected one, should be only a filled onse of the default. They dont exactly the same, so you can see some bad behaviours when clicking on this icon in the header. I edited the svg so they look now exactly the same, instead of the filling.


nowseemee:
Hi @niclas18, thanks for the PR! 🎉 🎉 🎉  

I'll discuss this change with our designers as soon as they are back from their vacation and I'll let you know.

niclas18:
Hello @nowseemee ,
do you have now an answer from the designers regarding to this issue?

PR #1183, opened 2022-08-24
test(visual): dry things up + better local reports

nowseemee:
@felix-ico can you have a look, please? I've removed some repetition from our visual test suites, tweaked the build script to run faster, and added `jest-html-reporter` to visualize test failures locally.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -8,11 +8,11 @@ jobs:
   prettier:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v2
 
       - name: Restore Lerna
         id: yarn-cache
-        uses: actions/cache@v3
+        uses: actions/cache@v2
         with:
           path: |
             node_modules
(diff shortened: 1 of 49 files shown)


felix-ico:
hi @nowseemee the code changes look good to me but when i run the tests locally get some tests that fail or time out. Am i supposed to do something different than before?

nowseemee:
Hi @felix-ico , thanks for the review. Which tests? Is it better with the current main branch?

felix-ico:
It's inconsistent but I'm getting a lot of failures
Screenshot 2022-08-25 at 15 45 20

Also when i check the generated html report the image links seems broken 

Screenshot 2022-08-25 at 15 44 08

Tried running on main too, and I am getting a lot of errors too, so I'm not sure now if it is a problem with my machine or docker set up.  In the past I've found that I get a lot better results with the --runInBand flag, I could try to run them again like that

felix-ico:
ok when i use the --runInBand flag locally only 2 tests fail (and it's one of those weird sub 0.1% difference in snapshots) so it seems all good.

felix-ico (approved):

PR #1186, opened 2022-08-26
chore: udpate fonts

felix-ico:


acstll (commented):
Only thing left is updating visual tests 🙌 (hopefully is not too painful with the current situation)

PR #1196, opened 2022-09-07
fix(dropdown-select-item): reflect `value` prop

### LINKED ISSUE #1194: DropdownSelectItem don't have a value in Vue 3 / Vite 3
**Scale Version**
@telekom/scale-components 3.0.0-beta.111

**Framework and version**
Vue 3.2.38
Vite 3.0.9 

**Current Behavior**
In our project the new DropdownSelect is broken, because the value attribute of DropdownSelectItem is not passed to the child components. In this case the Component looks really crazy when click on one option in the Dropdown.

**Expected Behavior**
Everything works like in the Storybook.
[Storybook-DropdownSelect](https://telekom.github.io/scale/?path=/docs/beta-components-dropdown-select--standard)

**Code Reproduction**
I use the same code, as in the Storybook:
`<scale-dropdown-select label="select">`
`<scale-dropdown-select-item value="caspar">Caspar</scale-dropdown-select-item>`
`<scale-dropdown-select-item value="cedric">Cedric</scale-dropdown-select-item>`
`<scale-dropdown-select-item value="cem">Cem</scale-dropdown-select-item>`
`</scale-dropdown-select>`

The CodeSandbox don't work at the moment with Vue 3 and Vite 3.

**Desktop (please complete the following information):**
 - OS: Windows 10 Enterprise - 21H2
 - Browser: Chrome
 - Version: 104.0.5112.102
 
**Additional context**
In the StoryBook I see with the Developer-Mode that all DropdownSelectItems have the value attribute in HTML-tree. In our project we have also the 3 DropdownSelectItems, but the value attribute don't passed into the HTML-tree.

For testing I add [here](https://github.com/telekom/scale/blob/main/packages/components/src/components/dropdown-select-item/dropdown-select-item.tsx) in line 16 a new line with the Host tag: `<Host value={this?.value}>` and end tag in line 28: `</Host>`.
Then it works like expected in our project.

acstll:
Fixes #1194

@nowseemee please take a look 🙏

### DIFF
--- a/packages/components/src/components/dropdown-select-item/dropdown-select-item.tsx
+++ b/packages/components/src/components/dropdown-select-item/dropdown-select-item.tsx
@@ -9,7 +9,7 @@ export class DropdownSelectItem {
   @Element() hostElement: HTMLElement;
   @Prop() selected?: boolean;
   @Prop() focused?: boolean;
-  @Prop() value?: any;
+  @Prop({ reflect: true }) value?: any;
 
   render() {
     return (


nowseemee (approved):
✨

acstll:
thank you @nowseemee, whenever you're ready 🚀

PR #1197, opened 2022-09-07
fix: add displayCloseButton prop, update storybook

maomaoZH:


packages/storybook-vue/stories/components/modal/Modal.stories.mdx
```diff
@@ -43,6 +43,13 @@ import ScaleModal from './ScaleModal.vue';
         disable: true,
       },
     },
+    scaleBeforeClose: {
+      description: '@deprecated in v3 in favor of kebab-case event names.',
+    },
+    'scale-before-close': {
+      description:
+        'It fires on every close attempt. Calling `event.preventDefault()` will prevent the modal from closing.',
```
acstll, line 51: Would it be possible to add a bit more info? something like this:

`event.detail.trigger` can be one of `CLOSE_BUTTON`, `ESCAPE_KEY` or `BACKDROP`, depending on the action that triggered the event.

packages/components/src/components/modal/modal.tsx
```diff
@@ -59,6 +59,8 @@ export class Modal {
   @Prop() closeButtonLabel?: string = 'Close Pop-up';
   /** (optional) title for close button */
   @Prop() closeButtonTitle?: string = 'Close';
+  /** (optional) display close button */
+  @Prop() displayCloseButton?: boolean = true;
```
acstll, line 63: What do you think about making the default `false`, thus changing the prop to something like `hideCloseButton` or `omitCloseButton`? I think "boolean" props should be false by default, I would like to have this by convention…

packages/components/src/components/modal/modal.tsx
```diff
@@ -274,18 +276,20 @@ export class Modal {
...
-                part="close-button"
-                onClick={() => this.emitBeforeClose('CLOSE_BUTTON')}
-                aria-label={this.closeButtonLabel}
-                title={this.closeButtonTitle}
-              >
-                <slot name="close-icon">
-                  <scale-icon-action-circle-close decorative />
-                </slot>
-              </button>
+              {this.displayCloseButton && (
```
acstll, line 279: For some reason the focus trap breaks when the button is not there… let's talk about this 🤓

acstll (changes requested):
More than changes I'm making questions 🙏 — thanks for the PR!

packages/storybook-vue/stories/components/modal/Modal.stories.mdx
maomaoZH: updated.

packages/components/src/components/modal/modal.tsx
maomaoZH: change to `omitCloseButton`

packages/components/src/components/modal/modal.tsx
maomaoZH: Hi, i tested again in our `modal.html`, and I also logged the focusable elements. I didnt encounter the issue though. 

https://user-images.githubusercontent.com/23239723/189601625-eace7edb-07f9-4590-825d-e758e400ce98.mov

acstll: I only had the issue in Storybook, but now it's also working there (just tested locally) 🎉 
image

packages/storybook-vue/stories/components/modal/Modal.stories.mdx
acstll: Thank you!

acstll:
All good! 👍

Issue #1198, opened 2022-09-08
Header logo issues depending on selected claim-lang


pomoalex:
**Scale Version**
3.0.0-beta.111

**Framework and version**
No framework

**Current Behavior**
When selecting an unsupported(but still valid) claim-lang, the logo disappears completely. Also, when selecting "en" the logo appears without the motto section, which wasn't the case previously.

**Expected Behavior**
I expect the main logo to be shown for any given locale(even if invalid) and the motto section of the logo to change in case of a recognized and supported locale.

**Code Reproduction**
```
<scale-app-shell id="app-shell">

    <scale-app-header slot="header" id="header" claim-lang="es">
    </scale-app-header>

</scale-app-shell>
```

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome
 - Version 105.0.5195.102

felix-ico:
Hi, Thank you for reporting this issue. As you can see I had opened a PR to achieve what you wanted, but after some discussion with the team I was able to gather that this is indeed the desired behavior of the brand header from a design/branding perspective.

PR #1201, opened 2022-09-09
fix: render logo even when claimLang is undefined

felix-ico:


felix-ico:
fixes https://github.com/telekom/scale/issues/1198

acstll:
I think we need to check with the design team for this one @felix-ico

felix-ico:
won't fix - the current implementation is correct after discussion with the team

PR #1203, opened 2022-09-09
chore: remove `web-animations-js` dep

acstll:
- Safari is no longer an issue, and [browser support is pretty great imo](https://caniuse.com/web-animation)
- `web-animations-js` is almost [15 kB gzipped](https://bundlephobia.com/package/web-animations-js@2.3.2) 🫣
- we're currently only using the Web Animations API in the Modal, and [it's implemented in a "progressive-enhancement" way](https://github.com/telekom/scale/blob/fdf09ccba8ce9a533ba7b8a1c31174a9f7bc377b/packages/components/src/components/modal/modal.tsx#L209-L224) (it won't animate but it'll open/close normally if the API is not supported)

@nowseemee 👀 🙏

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -68,12 +68,10 @@
   "dependencies": {
     "@duetds/date-picker": "1.2.0",
     "@floating-ui/dom": "^0.5.4",
-    "@proyecto26/animatable-component": "^1.1.8",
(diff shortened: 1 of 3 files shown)


nowseemee (approved):
🎉

PR #1204, opened 2022-09-12
feat: component ProgressBar -  add optional input property 'percentage_start'

michael-koeller:
### Feature Request:

Property 'percentage_start' at component `<scale-progress-bar>`

### Context

We are working on a Telekom-internal project. We just startet an new UI application for the TROPOS team, using this really great component library.

### Use Case

We have a web page displaying a multi-step, partly-manual/partly-automated processing status. Our UI will keep polliing this page to reflect any server-side status changes. In this scenario it's quite annoying that the progress bar would alway restart it's animation from scratch.<br/>
Instead, we would rather set the previous percentage and thus only have nice animations on actual status changes.

### DIFF
--- a/packages/components/src/components/progress-bar/progress-bar.tsx
+++ b/packages/components/src/components/progress-bar/progress-bar.tsx
@@ -26,6 +26,8 @@ export class ProgressBar {
   @Prop() busy?: boolean = false;
   /** (required) Progress bar percentage */
   @Prop() percentage: number = 0;
+  /** (optional) Progress bar percentage to start the animation from (default: 0) */
+  @Prop() percentage_start: number = 0;
   /** @deprecated - (optional) Progress bar customColor */
   @Prop() customColor?: string;
   /** (optional) Progress bar stroke width */
@@ -71,10 +73,10 @@ export class ProgressBar {
     }
   }
 
-  transitions = (width: number) => `
+  transitions = (width: number, width_start: number) => `
     @keyframes showProgress {
       from {
-        width: 0;
+        width: ${width_start}%;
       }
       to {
         width: ${width}%;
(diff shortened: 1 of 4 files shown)


packages/components/src/components/progress-bar/progress-bar.tsx
```diff
@@ -26,6 +26,8 @@ export class ProgressBar {
   @Prop() busy?: boolean = false;
   /** (required) Progress bar percentage */
   @Prop() percentage: number = 0;
+  /** (optional) Progress bar percentage to start the animation from (default: 0) */
+  @Prop() percentage_start: number = 0;
```
acstll, line 30: We use camelCase for props, if you could change this, it should be good to merge 🤓

acstll (changes requested):
Thanks a lot @michael-koeller — sorry for the delay in the reply!

packages/components/src/components/progress-bar/progress-bar.tsx
michael-koeller: Sure, done.

acstll: Thank you!

packages/components/src/components/progress-bar/progress-bar.tsx
```diff
@@ -71,10 +73,10 @@ export class ProgressBar {
     }
   }
 
-  transitions = (width: number) => `
+  transitions = (width: number, width_start: number) => `
     @keyframes showProgress {
       from {
-        width: 0;
+        width: ${width_start}%;
```
acstll, line 79: Oh, variable names also need to be camelCased, sorry 🙏 — thanks why tslint is failing.

acstll (changes requested):
I think you also need to update the snapshots (e.g. `yarn test --spec --updateSnapshot` in packages/components) — thanks again!

packages/components/src/components/progress-bar/progress-bar.tsx
michael-koeller: OK, I see. While checking the tests, I also added the new property to the existing test case. 😉

acstll: Amazing, thank you!

PR #1221, opened 2022-09-22
refactor(slider): mobile 📱 

acstll:
Replaces #1174

- [x] multi-thumb aka range
- [x] deprecate props `trackSmall`, etc.
- [x] update styles (use part selector [attr~=value])
- [x] styles for Android and iOS
- [x] show ["hash marks"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#a_range_control_with_hash_marks)
- [x] helper text
- [x] storybook
- [ ] visual tests update 😵‍💫


reld:
Nice! 🥷

PR #1222, opened 2022-09-23
feat: add helper text and use in text field

felix-ico:


packages/components/src/components/helper-text/helper-text.tsx
```diff
@@ -0,0 +1,93 @@
...
+      return (
+        <scale-icon-alert-success
+          color="var(--telekom-color-text-and-icon-functional-success)"
+          size={11}
+        ></scale-icon-alert-success>
+      );
+    }
+    if (variant === 'neutral') {
+      return;
+    }
```
acstll, line 72: I'd think this is not useful here, I would either remove it or move it to the top, so it's the first check and we return early…

packages/components/src/components/helper-text/helper-text.tsx
```diff
@@ -0,0 +1,93 @@
...
+    | 'warning'
+    | 'danger'
+    | 'success' = 'informational';
+
+  renderHelperIcon() {
+    const variant = this.variant;
+    if (variant === 'informational' || variant === 'warning') {
+      return (
+        <scale-icon-alert-information
+          color={
```
acstll, line 45: I'd almost sure setting the `color` prop is not needed, because all icons take the color of text (`currentColor`) by default…

packages/components/src/components/helper-text/helper-text.tsx
```diff
@@ -0,0 +1,93 @@
...
+            variant === 'informational'
+              ? 'var(--telekom-color-text-and-icon-functional-informational)'
+              : 'var(--telekom-color-text-and-icon-functional-warning)'
+          }
+          size={11}
```
acstll, line 50: We can probably make this a "constant" (e.g. `ICON_SIZE`)…

packages/components/src/components/helper-text/helper-text.css
```diff
@@ -0,0 +1,36 @@
+.helper-text {
+  display: flex;
+  align-items: center;
+  flex-direction: row-reverse;
+  justify-content: flex-end;
+  font: var(--telekom-text-style-small-bold);
```
acstll, line 6: I would add a scoped var (e.g. `--font` in the :host) for this one.

acstll (changes requested):
Thank you very much @felix-ico — I'm suggesting a few changes below (or above I don't know haha)

acstll:
Also, tests are failing @felix-ico 👀

packages/components/src/components/helper-text/helper-text.css
felix-ico: also added color variables

PR #1239, opened 2022-10-04
refactor(tab-nav,-header,-pannel): bringing back size prop

marvinLaubenstein:


packages/components/src/components/tab-header/tab-header.css
```diff
@@ -34,12 +34,12 @@
...
   height: var(--tab-height);
-  padding-right: calc(var(--tab-spacing) / 2);
-  padding-left: calc(var(--tab-spacing) / 2);
-  padding-top: var(--telekom-spacing-unit-x4);
-  padding-bottom: var(--telekom-spacing-unit-x4);
+  padding-right: var(--telekom-spacing-unit-x4);
+  padding-left: var(--telekom-spacing-unit-x4);
+  padding-top: var(--tab-spacing);
+  padding-bottom: var(--tab-spacing);
+  font-size: calc(var(--tab-height) / 2.4);
```
acstll, line 41: Could we use a font size token here? This calc is a bit 🙈

packages/components/src/components/tab-header/tab-header.css
```diff
@@ -129,3 +129,13 @@
...
 }
+
+/* small size */
+
+.tab-header--small {
+  padding-right: calc(var(--tab-spacing-small) / 2);
+  padding-left: calc(var(--tab-spacing-small) / 2);
+  padding-top: var(--telekom-spacing-unit-x4);
+  padding-bottom: var(--telekom-spacing-unit-x4);
+  font-size: calc(var(--tab-height-small) / 3);
```
acstll, line 140: …same here? 🤓

packages/components/src/components/tab-header/tab-header.css
```diff
@@ -34,12 +34,12 @@
...
   align-items: center;
   height: var(--tab-height);
-  padding-right: calc(var(--tab-spacing) / 2);
-  padding-left: calc(var(--tab-spacing) / 2);
-  padding-top: var(--telekom-spacing-unit-x4);
-  padding-bottom: var(--telekom-spacing-unit-x4);
+  padding-right: var(--telekom-spacing-unit-x4);
+  padding-left: var(--telekom-spacing-unit-x4);
+  padding-top: var(--tab-spacing);
+  padding-bottom: var(--tab-spacing);
```
acstll, line 40: I think small should be the default, no? If you think it's OK, let's change it now and we can then roll back if necessary 🙏

acstll (changes requested):
Mostly about the font size, is there a reason to use `calc` and not a token directly?

Thanks for updating this one (once again!) 🤓

packages/components/src/components/tab-header/tab-header.css
marvinLaubenstein: It was requested to take large as default, since large is the optimized for mobile use case 🙂

acstll (approved):
🤘

Issue #1232, opened 2022-09-29
Two Accordions inside of each other ignore level-heading


Lalaluka:
**Scale Version**
`3.0.0-beta.113`
`3.0.0-beta.56`

**Framework and version**
`vue@3.2.0` with both vite and webpack (vue-cli)

**Current Behavior**
If we take this code:
```
<scale-accordion heading-level="11">
  <scale-collapsible>
    <span slot="heading">Leo integer malesuada nunc vel risus</span>
    <p>Freegan kinfolk</p>
    <scale-accordion heading-level="12">
      <scale-collapsible>
        <span slot="heading">Leo integer malesuada nunc vel risus</span>
        <p>Freegan kinfolk</p>
      </scale-collapsible>
    </scale-accordion>
  </scale-collapsible>
</scale-accordion>
```
Both `scale-collapsible` titles will have `aria-level="11"` 

**Expected Behavior**
I would expect that only the first `scale-collapsible` would have `aria-level="11"`  and the second one `aria-level="12"`.

**Code Reproduction**
Minimal reproducible example:
[Scale and Vue3](https://codesandbox.io/s/scale-components-vue3-forked-vlrgcm?file=/package.json)

**Desktop (please complete the following information):**
 - OS: MacOs
 - Browser: chrome
 - Version: 104.0.5112.79

PR #1240, opened 2022-10-04
fix: heading-level propagation (#1232)

Lalaluka:
I wrote a maybe dirty fix for #1232.
I appreciate any comments for a cleaner fix.


Lalaluka:
Closing in favor of #1243

PR #1241, opened 2022-10-05
fix(checkbox): rename private class property to avoid clashing with restricted name

nowseemee:
…estricted name


packages/components/src/components/checkbox/checkbox.tsx
```diff
@@ -70,7 +70,7 @@ export class Checkbox {
   /** @deprecated in v3 in favor of kebab-case event names */
   @Event({ eventName: 'scaleChange' }) scaleChangeLegacy: EventEmitter;
 
-  private id = i++;
+  private internalId = i++;
```
nowseemee, line 73:
```suggestion
  readonly private internalId = i++;
```

PR #1246, opened 2022-10-06
Fix/progress bar

felix-ico:


acstll (changes requested):
Sent you some comments already 🙃

PR #1274, opened 2022-10-10
fix: add HCM borders, add range ariatext

felix-ico:


packages/components/src/components/slider/slider.tsx
```diff
@@ -379,9 +378,9 @@ export class Slider {
                         role="slider"
                         id={this.sliderId + '-from'}
                         aria-valuemin={this.min}
-                        aria-valuenow={this.value}
+                        aria-valuenow={`${this.valueFrom} to ${this.valueTo}`}
```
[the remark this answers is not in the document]
acstll: definitely… we'll need this — I remember some of these props to be a bit confusing but I'm not sure, we have to check it out.

Besides that… I think the [`aria-valuenow` should contain only the number](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/slider_role#associated_roles_states_and_properties).

According to [this on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuenow), `aria-valuetext` should be used for these cases @felix-ico:

> When the value to be announced, either the actual value or the value as a percent, may not be clear to users, the [aria-valuetext](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuetext) attribute should be used to provide a user-friendly representation of the value. When set, the valuetext string is announced instead of the valuenow numeric value. For example, if a slider represents the days of the week, so the day of the week's aria-valuenow is a number, the aria-valuetext property should be set to a string that makes the slider value understandable, such as "Monday".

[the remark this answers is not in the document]
felix-ico: ah nice, just checked the rating-stars implementation we can do it like that here too. I will revert aria-valuenow as well

packages/components/src/components/slider/slider.tsx
```diff
@@ -379,9 +387,9 @@ export class Slider {
...
                         aria-valuemax={this.max}
-                        aria-valuetext={`${this.value}`}
+                        aria-valuetext={`${this.valueFrom} to ${this.valueTo}`}
```
nowseemee, line 392:
```suggestion
                        aria-valuetext={this.getRangeAriaLabel()}
```

acstll: I'm going to sound picky but I would name the prop `innerAriaValueText` and thus the helper function `getAriaValueText()`. Mainly because an `aria-label` is a different and specific kind of attribute; and also because it's not really a translation but rather the actual value you're setting in whatever language. (This wouldn't be consistent with other components, but we're not consistent a 100% yet either, and we'll definitely work on this in the near future.)

@felix-ico and @nowseemee do you think?

felix-ico: updated it, was a mistake on my part.

acstll (approved):

Issue #1152, opened 2022-07-27
Add inputmode support to Text Field component


pomoalex:
**Describe the feature request**
It would be great to have the possibility to select the inputmode of a Scale Text Field component.

**Describe the use case**
The benefit of this would be displaying the correct keyboard to a mobile device user, for example displaying only the number keys in case of a PIN or telephone number.

**Describe alternatives you've considered**
The workaround I used so far is to manually find the input after the component has loaded and add the attribute.

marvinLaubenstein:
Good morning, in the upcoming releases the input mode for the text field should be included. If something is still missing, feel free to reopen the issue.

PR #1333, opened 2022-10-31
Feature/text field input mode

marvinLaubenstein:
Integrating Feature Request: 

- #1152


nowseemee (changes requested):
This should be [input type](https://www.w3schools.com/tags/att_input_type.asp) - and it's already supported. Am I missing something?

marvinLaubenstein:
@nowseemee the [input mode](https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/) is only about the keyboard, which is provided for the text field. 
I should have named the prop inputMode, but somehow it didn't work, I get an error message from Stencil.

nowseemee:
Ah I see, thanks for explaining :)

nowseemee (approved):

PR #1334, opened 2022-10-31
Fix/helper text ids

marvinLaubenstein:
### **Preventing/Solving problems with helper-text id**

**Affected components:**

- text-field
- dropdown
- dropdown-select
- radiobutton
- radiobutton-group
- checkbox
- input
- textarea
- date-picker


marvinLaubenstein:
Hi @acstll, it would be great if you could look over it, since you already know the topic and know what it's about. 

Thank you 🙂

nowseemee:
@acstll can you review this, please?

acstll:
Will do!

packages/components/src/components/checkbox/checkbox.tsx
```diff
@@ -150,7 +150,7 @@ export class Checkbox {
       return (
         <div
           part="helper-text"
-          id={text.id}
+          id={text.internalId}
```
acstll, line 153: I think this is a bug, `text.id` seems correct. (Honestly, I would avoid passing anything to the `renderHelperText` method and get the data needed directly in the method itself.)

acstll (changes requested):
I think this is much better than before, but still I guess there's room for duplicated ids within an app (e.g. there could be 2 simultaneus `helper-text-1`). Do you think we can use an util function like [this one in `shoelace`](https://github.com/shoelace-style/shoelace/blob/next/src/internal/auto-increment.ts)?

packages/components/src/components/checkbox/checkbox.tsx
marvinLaubenstein: Oh yes, that's right, I unfortunately slipped in between

marvinLaubenstein:
@acstll I have changed the way now so that the ids were generated by the utils file. I also checked it and could not find any id duplications. Great idea, thanks 🙂

packages/components/src/components/textarea/textarea.tsx
```diff
@@ -107,10 +105,11 @@ export class Textarea {
   @State() hasFocus: boolean = false;
 
   private focusableElement: HTMLElement;
+  private readonly internalId = generateUniqueId();
 
   componentWillLoad() {
     if (this.inputId == null) {
-      this.inputId = 'input-textarea' + i++;
+      this.inputId = 'input-textarea' + this.internalId;
```
acstll, line 112: I would add a dash after "textarea" for consistency (thanks for adding it in the other components) — the rest is ace 👌

acstll (approved):

packages/components/src/components/textarea/textarea.tsx
marvinLaubenstein: Oh yes, thank you 🙂

Issue #1233, opened 2022-09-30
Not able to hide default html tooltip from Telekom logo


VictoriaStefanikova:
**Framework and version**
Angular 14

**Code Reproduction**
Hi. I am currently using scale-tooltip component over the main Telekom logo in top left corner. The issue I am facing is, that while the tooltip is working as expected, I can see other html tooltip as well. I can not remove it with setting  logo-title attribute to empty string.

Can you suggest some workaround or working solution? Thank you.

![image](https://user-images.githubusercontent.com/56870904/193250393-ef35bad8-8ccf-4028-87a2-324b9ba2b4e6.png)

![image](https://user-images.githubusercontent.com/56870904/193250445-0ef33f91-6481-4cfc-a844-821c091a44e5.png)

marvinLaubenstein:
Hi @VictoriaStefanikova, in the next releases there will be a possibility to hide the logo title with the prop `logoHideTitle`. 
If there is still a problem after that please don't hesitate to reopen the ticket.

PR #1335, opened 2022-10-31
Feat/logo title hide

marvinLaubenstein:
Integrating Feature to solve "Bug"-Issue:

- #1233


nowseemee:
nice! thanks :)

packages/components/src/components/telekom/logo/logo.tsx
```diff
@@ -46,6 +46,9 @@ export class Logo {
   @Prop() styles?: string;
   @Prop() focusable: boolean = true;
   @Prop() scrollIntoViewOnFocus: boolean = false;
+  /** (optional) Hide all logo related titles */
+  @Prop() hideTitle?: boolean = false;
```
nowseemee, line 50: hi @marvinLaubenstein thanks a lot for the PR! I think it would make sense to add this prop to `app-shell` and `app-header` and pipe it down to the logo so the users of the header can also benefit from this. Also, let's call it `logoHideTitle` everywhere.

nowseemee (changes requested):
see comment

packages/components/src/components/telekom/logo/logo.tsx
marvinLaubenstein: Sounds like a very good idea, I change that.

nowseemee (approved):
💫

Issue #1209, opened 2022-09-13
header expands over entire page when page width is between 1023px and 1024px


JoKurth:
**Scale Version**
v3.0.0-beta.112

**Framework and version**
Vue 3.2.38
Vite 3.1.0

**Current Behavior**
The magenta part of the header expands over almost the entire page when the size of the page/window is between 1023px and 1024px. It seems like the corresponding media-queries should not be `max-width: 1023px` and `min-width: 1024px` but rather `max-width: 1023.99px` and `min-width: 1024px`
![image](https://user-images.githubusercontent.com/71062996/189927546-ab44cd5e-ccb6-4f69-9668-27aff85f8a19.png)

**Expected Behavior**
The header should break smoothly and not hide the page.

**Code Reproduction**
Use the `<scale-app-shell>`. I tested the issue in the following codesandboxes by simply adding the `scale-app-shell` around the given code.
[Scale and React18](https://codesandbox.io/s/scale-components-react18-75ci1x)
[Scale and Vue3](https://codesandbox.io/s/scale-components-vue3-b14993)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Brave (Chromium)
 - Version 1.43.89 (Chromium: 105.0.5195.102)

tshimber:
hi! not it reproduces on 1039.20 px. should i create another issue?
![image](https://github.com/telekom/scale/assets/110041439/d847cb58-1246-4899-9a50-f86d66d373f4)

felix-ico:
hi @JoKurth @tshimber, I just tried testing this with the new header and could not reproduce the issue (see attached video).


https://github.com/telekom/scale/assets/97440128/af727df5-da3b-4968-ae16-e683e0e73f86



Please reopen the issue if you think the header is still buggy, perhaps I misunderstood the issue.

tshimber:
@felix-ico hi) we use old app-shell, and honestly nobody cares except me) i guess we need to move to new one if somebody will complain, so nothing from my side here, thanks!

PR #1339, opened 2022-10-31
fix(app-header): solving issue with width between 1023 and 1024 px

marvinLaubenstein:
Solving "Bug"-Issue:

- #1209


nowseemee:
How does one set the viewport between 1023 and 1024? 🃏

marvinLaubenstein:
By manually moving the window and then landing the width at, for example, 1023.3 🤓. With the comma pixels there are no more problems. Even if it is actually illogical to assign comma pixels

nowseemee:
Hm, I'm not able to reproduce the original issue but with the changes in this PR I see this:
image

nowseemee (changes requested):
visual tests should not show any change. Seems like there is a problem when the viewport is 1024px.

marvinLaubenstein:
I could reproduce it again with the current main branch

Bildschirmfoto 2022-11-28 um 21 07 51

felix-ico:
from the discussion with devs: maybe we can try to expand the media query instead of adding fractions of pixels. for example: `min-width: .... AND max-width: ...`

felix-ico:
we're closing this, it's fixed in https://github.com/telekom/scale/pull/1650

Issue #915, opened 2022-03-02
FlyoutItem containing `a` isn't accessable with keyboard


christopherGdynia:
Hello there,

My Flyout is containing an `anchor` `tag`, but selecting it with `Space` won't trigger the `anchor` `tag`.

```tsx

import {
  ScaleButton, ScaleIconActionLink, ScaleIconActionMore, ScaleMenuFlyout, ScaleMenuFlyoutDivider, ScaleMenuFlyoutItem, ScaleMenuFlyoutList
} from '@telekom/scale-components-react';
import { forwardRef, ReactNode } from 'react';

const BaseFlyout = forwardRef<
  HTMLScaleMenuFlyoutElement,
  { url: string; name: string; children: ReactNode }
>((props, ref) => (
  <ScaleMenuFlyout ref={ref}>
    <ScaleButton slot="trigger" iconOnly variant="ghost" aria-label="Kontextmenü öffnen">
      <ScaleIconActionMore />
    </ScaleButton>
    <ScaleMenuFlyoutList>
      <ScaleMenuFlyoutItem>Portal {props.name}</ScaleMenuFlyoutItem>
      <ScaleMenuFlyoutDivider />
      {props.children}
      <ScaleMenuFlyoutItem data-link>
        <ScaleIconActionLink slot="prefix" />
        <a href={props.url} target="_blank" rel="noreferrer" className="link">
          Öffnen
        </a>
      </ScaleMenuFlyoutItem>
    </ScaleMenuFlyoutList>
  </ScaleMenuFlyout>
));

export default BaseFlyout;

```

With workaround
```tsx

import { ScaleIconActionRemove, ScaleMenuFlyoutItem } from '@telekom/scale-components-react';
import { useEventListenerFactory } from 'lotus-hooks';
import { useSelectedPortalsStore, useSetSelectedPortals } from 'src/hooks';
import BaseFlyout from '../shared/baseFlyout';
export interface IFlyoutProps {
  name: string;
  url: string;
}

const Flyout = ({ name, url }: IFlyoutProps) => {
  const selectedPortals = useSelectedPortalsStore();
  const setSelectedPortals = useSetSelectedPortals();

  const ref = useEventListenerFactory<HTMLScaleMenuFlyoutElement>(
    'scale-select' as any,
    (e: any) => {
      if (e.detail.item.getAttribute('data-remove')) {
        const newState = [...selectedPortals];
        newState.splice(newState.indexOf(name), 1);
        setSelectedPortals(newState);
[...]

PR #1367, opened 2022-11-16
feat(menu-flyout): pass `originalEvent` in `scale-select` event

acstll:
Related to https://github.com/telekom/scale/issues/915#issuecomment-1310460002

> We can include the original click event that has ctrlKey inside the details of the emitted scale-select event. That should be an easy fix. We could name it originalEvent (in a jQuery fashion).

@felix-ico could you please review this one 🙏


felix-ico (commented):
Looks good to me, just like in the original conversation I wonder

> I am just wondering, where this could be usefull aswell :D

felix-ico (approved):

PR #1368, opened 2022-11-18
Update examples

felix-ico:


felix-ico:
@acstll could you have a look at this when you have some time? I updated the dependencies for all out example projects and made them consistent re: font-styles and using the neutral scale package

marvinLaubenstein (approved):
Looks good to me🤓

PR #1370, opened 2022-11-18
feat(brand-bar): add new component

nowseemee:
@acstll this is ready


acstll (approved):
Hey, this is a great start… 🌈 (thank you).

I have a few comments (below), but we can address them moving forward… we can merge now.

- I would prefer to not have function props (in general) e.g. `appNameClick` but this we can discuss — if there's a "better" way
- having `pageYOffset` as State triggers a `render()` call on each `scroll` event, maybe this can hurt performance? (maybe the logic could be simplified a bit?)
- I think we don't need the `telekom-app-name` component anymore, we could have the styles in `telekom-header`
- I will have suggestion for slot names afterwards 😛

Issue #1375, opened 2022-11-22
Button form attribute


nikolab-pn:
**Describe the feature request**
Scale button component doesn't pass form attribute to button component

**Describe the use case**
This attribute lets you associate button elements to forms anywhere in the document, not just inside a form.

Issue #1366, opened 2022-11-16
Passing an aria-label attribute to the scale-button


equitable66:
Due to the fact that we have to ensure accessibility, we miss the attribute aria-label for the scale-button, so that an alternative text to the title attribute can be read out by a screen reader.
Passing an aria-label attribute to the scale-button would  add it to the associated shadow button element .

PR #1376, opened 2022-11-22
feat(button): aria-label and form attribute

marvinLaubenstein:
- **Aria-label attribute** (#1366): Has been tested for button and a-tag and is compliant
- **Form attribute** (#1375) : Is only allowed for button tag, so the prop would only be possible without href. Would have to communicate somehow or discard the feature.


marvinLaubenstein:
Hey @acstll 🙂, could you look over it if this fits so far or if you have change suggestions

marvinLaubenstein:
Hey @acstll 🙂, could you look over it if this fits so far or if you have change suggestions

packages/components/src/components/button/button.tsx
```diff
@@ -61,6 +61,10 @@ export class Button {
   @Prop() download?: string;
   /** (optional) Set `tabindex` in the inner button or link element */
   @Prop() innerTabindex?: number;
+  /** (optional) Button aria-label */
+  @Prop() ariaLabel?: string;
```
acstll, line 65: I think this prop should be renamed to `innerAriaLabel` but I'm not sure. Because `aria-label` can practically be added to anything and `scale-button` is indeed an element, now you would get 2 aria-labels instead of the one you actually want in the actual button inside the shadow DOM. Would you agree?

packages/components/src/components/button/button.tsx
```diff
@@ -61,6 +61,10 @@ export class Button {
...
+  /** (optional) Button form */
+  @Prop() form?: string;
```
acstll, line 67: Regarding the `form` prop, it should work as it is now, but I do have 2 suggestions:

- we have [a couple](https://github.com/telekom/scale/blob/ee568c01e482508f1835f6f6d46b9595d86822b5/packages/components/src/components/button/button.tsx#L133) [of "hacks"](https://github.com/telekom/scale/blob/ee568c01e482508f1835f6f6d46b9595d86822b5/packages/components/src/components/button/button.tsx#L96) in place to deal with forms already. In them we do `this.hostElement.closest('form')` to query for a form. If you use the `form` attribute, probably the button is not a direct child of the relevant form, so we would need to query for that `id` instead of just a <form> if the `form` prop is present… to make sure the hack works as expected…
- also, if we add the `form` prop, I think it would make sense to also add the rest of the `form-` attributes a button can take — in which case I think they would need to be copied over to the [hack button](https://github.com/telekom/scale/blob/ee568c01e482508f1835f6f6d46b9595d86822b5/packages/components/src/components/button/button.tsx#L104-L106) too.
  - [formaction](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formaction)
  - [formenctype](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formenctype)
  - [formmethod](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formmethod)
  - and so on…

acstll (changes requested):
Thank you @marvinLaubenstein (and sorry for the delay). Please let me know what you think (or if you have any doubts).

packages/components/src/components/button/button.tsx
marvinLaubenstein: Thanks for the tip 🙌🏻, I would be in favor of removing the form property. Instead I will offer an alternative to the issue creator.

felix-ico:
I think this one is also related https://github.com/telekom/scale/issues/1405

acstll (approved):

Issue #1372, opened 2022-11-21
Scale-tab-nav always graps tab nav


maiktal:
**Scale Version**
3.0.0-beta.117

**Framework and version**
Angular 14

**Current Behavior**
The tab navigation component graps the focus automatically. The prevents e.g. SinglePage Apps from setting the focus to other elements on page navigation.

**Expected Behavior**
The component should not grap the autofocus per default. Introduce a prop to enable auto focus.

**Code Reproduction**
-

**Desktop:**
- OS: Windows 10
- Browser CHROME
- Version 107.0.5304.107 (Offizieller Build) (64-Bit)

**Additional context**
This is a major issue for us. We tested it with a blind person. The person had no chance to "see" that the page changed.

sulyok-daniel:
The issue still exists.
When rendering a `scale-tab-nav` for the first time (e.g. after a hard reload), all is fine. But in every subsequent rendering the `scale-tab-nav` sets focus on the previously selected tab.

Since we have an SPA, we set focus manually on the H1 element after a page load, so screen reader users understand that a navigation took place and know where they are in our application. scale-tab-nav's current behaviour prevents this.

sulyok-daniel:
Now, in Scale 3.0.0-beta.130, the initial issue (tab grabbing focus automatically after the page loads) is resolved, thank you! :)

But I regret there is a new related issue:
Now, when switching between the tabs via keyboard LEFT/RIGHT arrows, the selection of the tab header is not updated, so screen reader users are not informed about the changes.

**current behaviour**
After RIGHT click from 1st tab, **tab-header 1** remains selected, even though **tab-header 2** is active and **tab-panel 2** is displayed.
![image](https://user-images.githubusercontent.com/2097315/231458326-0aed225d-1f35-4c46-8c51-cfb9c0f6423f.png)

**expected behaviour**
After RIGHT click from 1st tab, **tab-header 2** should be selected, as was in scale v 3.0.0-beta.126:
![image](https://user-images.githubusercontent.com/2097315/231458977-79ea7df2-849d-4940-b9bd-877d4d40f57e.png)

acstll:
@sulyok-daniel thanks for the patience! We just merged https://github.com/telekom/scale/pull/1729 with a final fix 🤞 — releaes will follow…

sulyok-daniel:
It works fine, the issue is gone! Thank you! 👍

PR #1383, opened 2022-11-24
fix: prevent focus on first render

felix-ico:


felix-ico:
fixes https://github.com/telekom/scale/issues/1372.

not sure if the new prop `autoFocus` should be true or false by default? If we want to keep the previous behavior (first element focussed) I should change it to `true`, if we want it to behave as the issue reporter wishes we can keep it with `false` as it is implemented here.

felix-ico:
@acstll could you review this?

acstll (changes requested):
Hey @felix-ico thanks for taking care of this one! It's working as expected now 👍 …but I think we shouldn't be exposing a `firstRender` attribute in Tab Header, it doesn't seem useful for consumers 🤔 — Could we make this something internal?

felix-ico:
@acstll i removed the stencil prop and used a regular attribute instead, seems to work as expected.

acstll (approved):
👌

Issue #1371, opened 2022-11-21
Configurable Screen Reader alt-text for "open profile menu"


maiktal:
**Scale Version**
3.0.0-beta.117

**Framework and version**
Angular 14

**Current Behavior**
The screen reader reads only the name displayed on profile menu. 

**Expected Behavior**
An alternative Text for the screen reader with the intention to open an profile menu for the user
can be configured at the component.

**Code Reproduction**
See the picture.

![image](https://user-images.githubusercontent.com/118713833/202983168-4e2f5009-393a-47de-b2ec-32364782367d.png)

**Desktop:**
 - OS: Windows 10
 - Browser CHROME
 - Version 107.0.5304.107 (Offizieller Build) (64-Bit)

**Additional context**
For us its a major problem because for SR users this feature is unusable! (we tested it with a blind person)

felix-ico:
closing this issue as it refers to the deprecated brand header. As always feel free to reopen if it's still relevant to you

PR #1387, opened 2022-11-28
Refactor/brand header

marvinLaubenstein:


felix-ico:
https://github.com/telekom/scale/issues/1371

felix-ico (changes requested):
@marvinLaubenstein I think there is a piece missing here, the appShell doesn't pass the useMenuAriaLabel to the app-header, or am i missing something?

Also I guess it would be nice to have this feature for all navigation icons but for now we could just pass it for the user menu 🤓 - let's think about it for the new telekom header

felix-ico (approved):

felix-ico (approved):
LGTM 🚀

Issue #1257, opened 2022-10-07
Data Grid sorting does not work as expected


mato-a:
**Scale Version**
v3.0.0-beta.114 - latest at the time of writing

**Framework and version**
Any

**Current Behavior**

Clicking on the sort button in Data Grid table header will switch between 3 states:

- sorted in ascending order
- sorted in descending order
- unsorted - UI doesn't show this in any special way - either sorted by ascending or descending order icon is shown (small arrow-like icon pointing up or down)

**Expected Behavior**

Clicking the sort buttons in Data Grid table header should switch between 2 states:

- sorted in ascending order
- sorted in descending order

**Code Reproduction**

Can be reproduces also in Storybook: https://telekom.github.io/scale/?path=/docs/components-data-grid--standard

nowseemee:
is this really a bug we should fix?

felix-ico:
@nowseemee maybe we need some UX feedback but I thought that sorting between three different states (ascending/descending/no sorting) is unexpected behavior.

nowseemee:
@mato-a good point, thanks for posting an update - we should discuss this within the team together with @felix-ico 🙌🏻

felix-ico:
hi @mato-a this should be fixed in the next release https://github.com/telekom/scale/pull/1396

PR #1396, opened 2022-11-30
fix: only allow sorting ascending and descending

felix-ico:
this should fix https://github.com/telekom/scale/issues/1257

### DIFF
--- a/packages/components/src/components/data-grid/data-grid.tsx
+++ b/packages/components/src/components/data-grid/data-grid.tsx
@@ -392,7 +392,7 @@ export class DataGrid {
         ? 'ascending'
         : sortDirection === 'ascending'
         ? 'descending'
-        : 'none';
+        : 'ascending'
     this.fields[columnIndex].sortDirection = newSortDirection;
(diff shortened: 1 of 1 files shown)


nowseemee:
to me this thing looks much more like a feature than a bug

acstll:
I'm in favor of keeping the unsorted option, taking into account [`aria-sort="none"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-sort#values) is the default state. Also following @mato-a suggestion:

> I am OK with unsorted to be considered a 3rd possible state, just please make it also apparent to the user in the UI. 🙂

This is how I would solve this. @felix-ico what do you think? This removes the visible arrow (when not hovered!) when `aria-sort` is none.

image

acstll (changes requested):
Please see https://github.com/telekom/scale/pull/1396#issuecomment-1473943917 🤓

felix-ico:
> This removes the visible arrow (when not hovered!) when aria-sort is none.

sounds good to me! so I guess no need for a further UX review on this?

Issue #1358, opened 2022-11-10
Add autocomplete-Prop to TextField


niclas18:
**Describe the feature request**
Unfortunately, there is no property for "autocomplete" in the Scale-TextField component. I ask that you add this so that the appropriate options can be set and correctly recognized by the browser.

**Describe the use case**
"The HTML autocomplete attribute lets web developers specify what if any permission the user agent has to provide automated assistance in filling out form field values, as well as guidance to the browser as to the type of information expected in the field."
"The source of the suggested values is generally up to the browser; typically values come from past values entered by the user, but they may also come from pre-configured values. For instance, a browser might let the user save their name, address, phone number, and email addresses for autocomplete purposes. Perhaps the browser offers the ability to save encrypted credit card information, for autocompletion following an authentication procedure."
Source: [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)

**Describe alternatives you've considered**
In my opinion, there are no alternatives to setting the property.

**Additional context**
The problem is that modern browsers try to use the "ID" and "Name" of the input element to derive which values are expected there (e.g. for street or city name). Chrome then inserts these without first alerting the user to the possibility of a "false detection". The only way to ensure correct insertion is to set a hard-defined value for "autocomplete" (see mdn docs). In addition, without this property, it is also not possible to completely prevent automatic insertion. In our project, however, it is a requirement that in the case of security-critical changes (which require entering a password), the user must enter his password himself and this does not happen automatically from the browser.

niclas18:
Clossed because of #1400

PR #1400, opened 2022-12-02
add autocomplete-property to Scale-TextField

niclas18:
close my issue #1358


packages/components/src/components/text-field/text-field.tsx
```diff
@@ -105,6 +105,8 @@ export class TextField {
   @Prop() list?: string;
   /** (optional) the input should automatically get focus when the page loads. */
   @Prop() inputAutofocus?: boolean;
+  /** (optional) custom value for automplete HTML attribute */
+  @Prop() autocomplete?: string;
```
nowseemee, line 109: can you rename it to `inputAutocomplete`, please? So it won't clash with the native attribute name.

packages/components/src/components/text-field/text-field.tsx
```diff
@@ -105,6 +105,8 @@ export class TextField {
   @Prop() list?: string;
   /** (optional) the input should automatically get focus when the page loads. */
   @Prop() inputAutofocus?: boolean;
+  /** (optional) custom value for automplete HTML attribute */
```
nowseemee, line 108:
```suggestion
  /** (optional) custom value for autocomplete HTML attribute */
```

packages/components/src/components/text-field/text-field.tsx
niclas18: Thanks for your feedback, I changed/add it in my last [commit](https://github.com/telekom/scale/pull/1400/commits/ea1b4a868bc70ad7ae6b0bd0863643a658967ccc).

packages/components/src/components/text-field/text-field.tsx
niclas18: Thanks for your feedback, I changed/add it in my last [commit](https://github.com/telekom/scale/pull/1400/commits/ea1b4a868bc70ad7ae6b0bd0863643a658967ccc).

nowseemee:
Many thanks for your contribution, @niclas18! We'll release it early next week.

PR #1411, opened 2022-12-06
fix: add localization to flyout menu of data grid

maomaoZH:


felix-ico:
@felix-ico provide review

PR #1430, opened 2022-12-16
fix: add disabled link option

felix-ico:


felix-ico:
@acstll could you check this one? also re: the above comment

packages/components/src/components/button/button.tsx
```diff
@@ -198,19 +201,22 @@ export class Button {
...
-            part={basePart}
-            tabIndex={this.innerTabindex}
-          >
-            <slot />
-          </a>
-        ) : (
+            <a
+              ref={(el) => (this.focusableElement = el)}
+              class={this.getCssClassMap()}
+              href={this.href}
```
acstll, line 207: I would make this `href={this.disabled ? null : this.href}`, effectively the same as removing the attribute from the host without touching the host, which would be impolite…

packages/components/src/components/button/button.tsx
```diff
@@ -111,6 +111,9 @@ export class Button {
   connectedCallback() {
     this.setIconPositionProp();
     this.appendEnterKeySubmitFallback();
+    if (this.disabled) {
+      this.hostElement.removeAttribute('href');
+    }
```
acstll, line 116: This can be declaratively, and also removing an attribute set by the user might be a bit rude haha, there's a suggestion below.

packages/components/src/components/button/button.tsx
```diff
@@ -198,19 +201,22 @@ export class Button {
...
+              ref={(el) => (this.focusableElement = el)}
+              class={this.getCssClassMap()}
+              href={this.href}
+              download={this.download}
+              target={this.target}
+              rel={this.target === '_blank' ? 'noopener noreferrer' : undefined}
+              part={basePart}
+              tabIndex={this.innerTabindex}
+              role="link"
+              aria-disabled={this.disabled}
```
acstll, line 214: This needs to be `aria-disabled={this.disabled ? 'true' : null}` so it actually render the "true" value, otherwise is not valid I think.

packages/components/src/components/button/button.tsx
```diff
@@ -238,7 +244,7 @@ export class Button {
       !this.iconOnly &&
         this.iconPosition &&
         `button--icon-${this.iconPosition}`,
-      this.disabled && !this.href && `button--disabled`
```
acstll, line 241: This is good 👍

acstll (changes requested):
Hey, I wouldn't be in favor of allowing disabled links but since it's doable without much hassle, I think we can. Also [the article you posted](https://www.scottohara.me/blog/2021/05/28/disabled-links.html) is great 🙌.

You almost had it, I'm suggesting a few changes only.

packages/components/src/components/button/button.tsx
felix-ico: true, removed in favor of your implementation

Issue #1464, opened 2023-01-10
Data Grid access to toggle check all programmatically  


Fanky9:
**Scale Version**
3.0.0-beta.117

**Framework and version**
vue 3.2.45

![image](https://user-images.githubusercontent.com/115473672/211551881-da0d8ccb-2a5f-4cae-bd99-6f7fd8a1799c.png)

**Additional context**
Hi i am using scale-data-grid and I have feature to remove some items from table and i have problem to uncheck this checkbox. is there any change to access that checkbox and uncheck it? Or run some validation after rows changes and revalidate checkAll flag?

acstll:
In order to access DOM element for the checkbox, one could do:

```js
// assuming `el` is the <scale-data-grid> we want to target
const selectAllCheckbox = el.shadowRoot.querySelector('scale-checkbox[aria-label="Select"]')
selectAllCheckbox.checked = false
```

This is a bit "hacky" but does the trick. We could eventually add a method to make that easier?

@Fanky9 does this help?

<img width="646" alt="Captura de pantalla 2023-02-08 a les 15 46 56" src="https://user-images.githubusercontent.com/851105/217564619-3552576c-f965-4767-a7df-80d3d5dccc4f.png">

Fanky9:
@acstll thanks this workaround works :)

PR #1465, opened 2023-01-10
fix: update header checkbox when selectable data grid is updated

felix-ico:
this should fix https://github.com/telekom/scale/issues/1464

### DIFF
--- a/packages/components/src/components/data-grid/data-grid.tsx
+++ b/packages/components/src/components/data-grid/data-grid.tsx
@@ -187,7 +187,9 @@ export class DataGrid {
     this.addResizeObserver();
   }
   componentDidUpdate() {
-    this.selectable && this.updateReadableSelection();
+    if (this.selectable) {
+      this.updateReadableSelection();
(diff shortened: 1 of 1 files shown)


acstll:
I failed to see the difference in behaviour for the "select all" checkbox between this branch and `main`… I might be missing something… 😅

@felix-ico what is the exact fix this change does?

acstll:
I also added a comment in there https://github.com/telekom/scale/issues/1464#issuecomment-1422722876 based on my understanding of the issue… might be wrong though.

felix-ico:
>I failed to see the difference in behaviour for the "select all" checkbox between this branch and main… I might be missing something… 😅

Oh it's been a while but from what i remember it was about removing/adding grid rows programmatically/dynamically. I think what happened was that if a selected/checked row was removed (programmatically) the main checkbox (in the table head) would remain checked, even if all the other rows were unchecked. 

I guess we can close this PR though, if the workaround you provided does the trick

PR #1468, opened 2023-01-10
feat(telekom-nav-flyout): add default x spacing to match header container

acstll:


acstll:
@nowseemee please 👀 🙏

PR #1477, opened 2023-01-16
feat(telekom-header): consolidate slots, improve markup

acstll:
- [x] consolidated slot names (including proposal for mobile)
- [x] simplified markup
- [x] improved styles for app name (no extra component)
- [x] start work on decoupling mobile menu from mobile flyout (`…flyout-canvas`)


nowseemee (commented):
LGTM!

Issue #1446, opened 2023-01-04
Accessibility of error text in ScaleTextField


tschinski:
Hi guys,

we're facing an accessibility issue with ScaleTextField component.
Unfortunately the error text of an ScaleTextField is not recognized by screen readers. 

https://telekom.github.io/scale/?path=/docs/components-text-field--with-error

Actually, all kind of helper texts are not recognized.

Do we miss something?

Thanks in advance and best regards
Tobias

marvinLaubenstein:
Hi @tschinski, sorry that I have not answered for so long. The accessibility errors and/or deficits should be fixed in the near future.

Issue #1426, opened 2022-12-16
Scale-text-field - error messages not read by Screen readers


wgmaik:
**Scale Version**
3.0.0-beta.117

**Framework and version**
Angular 14

**Current Behavior**
Not possible to set aria-describedby & aria-details attribute on input fields in order to link input to error messages.

![image](https://user-images.githubusercontent.com/10576372/208058348-1834a0fb-5255-4568-b828-109e2bb1b517.png)

**Expected Behavior**
Set aria-describedby & aria-details attribute on input fields in order to link input to error messages. We need this to link custom error messages as displayed in the image to enable display of a bunch of error messages. Currently its only possible to display a single error message. A list would make it more flexible and usable!

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-details
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby

**Desktop (please complete the following information):**
- OS: Windows 10
- Browser CHROME
- Version 107.0.5304.107 (Offizieller Build) (64-Bit)

marvinLaubenstein:
Hi @wgmaik, sorry that I have not answered for so long. The accessibility errors and/or deficits should be fixed in the near future.

georgwittberger-telekom-mms:
We also have the issue that error messages are not announced by screen readers. It seems like there are multiple things why this is not working.

![0AXVwtL6iB](https://github.com/telekom/scale/assets/64543811/ac89ac1c-7db8-4c86-8a00-dcc5810c4a4b)

- `scale-helper-text` is a sibling of the `div` having `aria-live="polite"` (not a child), therefore helper text is not announced when it appears.
- `input` element has `aria-invalid` attribute (maybe should have string value `"true"`) but is missing `aria-errormessage` which should refer to ID of `scale-helper-text` (see [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage) for details)

`aria-invalid="true"` and `aria-errormessage="id-of-helper-text"` should only be set on `input` element if field is rendered with prop `invalid`. In this case it could be reasonable to omit `aria-describedby` because it could make screen readers read the helper text twice.

PR #1487, opened 2023-01-18
Fix/helpertext text field accessibility

marvinLaubenstein:
In this PR the following issues are solved:
- #1426 
- #1446 

Details about the changes:
- added `aria-details` property to text-field component
- fixed text-field component `aria-describedBy` property


marvinLaubenstein:
Hi @felix-ico, would be great if you could look over it whether this can be merged like that 🙂

felix-ico (approved):
tested locally with a screenreader, LGTM

felix-ico:
also https://github.com/telekom/scale/issues/1572

PR #1533, opened 2023-01-27
fix: remove mobile heading on nested levels

felix-ico:


nowseemee (approved):

nowseemee:
what happens when it's a link and has children?

felix-ico:
we've decided against this

PR #1535, opened 2023-01-27
feat(telekom-header): data api

nowseemee:


packages/components/src/components/telekom/telekom-header-data-back-compat/telekom-header-data-back-compat.tsx
```diff
@@ -0,0 +1,384 @@
...
+                );
+              })}
+
+            <scale-telekom-nav-item class="burger-item">
+              <button>
+                <scale-badge>
+                  <scale-icon-action-menu></scale-icon-action-menu>
+                </scale-badge>
+              </button>
+              <scale-telekom-nav-flyout class="mobile-nav-flyout">
```
acstll, line 226: These `burger-item` and `mobile-nav-flyout` classes are 💩 because people will need the code but it's "hidden", no? Please don't do anything about them now. 

I'll try and add an update to `scale-telekom-header` now (probably some news props) so we can remove them (here and in the .vue file where they're defined).

acstll: The classes are my fault in the first place… 🫣

acstll (changes requested):
These is just ace. Thanks a lot @nowseemee — I'm requesting some small changes and leaving a note (regarding those classes).

nowseemee:
Thanks for the review @acstll, I've pushed a commit.

PR #1560, opened 2023-01-31
Fix/more header feedback

felix-ico:


packages/components/src/components/telekom/telekom-header/telekom-header.css
```diff
@@ -299,6 +299,12 @@ slot[name='functions'] {
   text-decoration: none;
 }
 
+@media screen and (--lg) {
+  [part~='app-name-text'] {
+    --font-size-app-name: var(--telekom-typography-font-size-small);
+  }
+}
+
```
acstll, line 307: Maybe it's better to just update the variable instead of assigning it to that specifc part…

```diff
@media screen and (--lg) {
-  [part~='app-name-text'] {
+ :host {
    --font-size-app-name: var(--telekom-typography-font-size-small);
  }
}
```

acstll (approved):
Please check the comment regarding the media query, other than that… 🚀

acstll:
@felix-ico also please check the prettier check (failing) 🙏

packages/components/src/components/telekom/telekom-header/telekom-header.css
felix-ico: ahh true

unrelated: i find this css is a bit because the variable actually gets used in the telekom-mobile-menu.css - not sure if we should change something there

acstll: me neither at this point haha…

PR #1563, opened 2023-02-02
fix(progress-bar): inner bar for percentage 0

### LINKED ISSUE #1500: progress bar is set 0 percentage, it still magenta color
**Scale Version**
latest version

**Framework and version**
In vue app. But i see it also in storybook

**Current Behavior**
Set the progress bar to 0 percentage. There is still magenta color appear.  (see screenshot)
Screenshot 2023-01-23 at 09 54 49

**Expected Behavior**
The progress bar should not have magenta color

I checked it was caused by `progress-bar__inner` class. It has border value `1px solid transparent`. When removing it, the progress bar will not have any filled color. Any reason why it is done like this? or is it a bug?

acstll:
Fixes #1500


acstll:
@marvinLaubenstein would you be so kind to review this one? 🙏

marvinLaubenstein (approved):
Looks good 🤓

marvinLaubenstein:
Nice, sounds good. 

The failures in the Brand Header snapshots looks like the ones we always have... I don't know why...

Issue #1199, opened 2022-09-08
Don't render expand button in data grid when HTML cell is empty


nikolab-pn:
**Describe the feature request**
Don't render expand button in data grid when HTML cell is empty

**Describe the use case**
We would like to show data as expanded row but not for all rows

https://codesandbox.io/s/zen-torvalds-y7o4x5?file=/src/App.js

felix-ico:
this should be available in the next release https://github.com/telekom/scale/pull/1573

PR #1573, opened 2023-02-06
fix: do not render expand button when html field is empty

felix-ico:
https://github.com/telekom/scale/issues/1199


acstll (approved):
This looks good! (…you only need to run the damn `prettier` 💅)

PR #1592, opened 2023-02-14
feat: add link target variables to menu and footer

niklaswa:
Hi there, we're using scale in one of the web applications at TelekomCLOUD marketplace and missing a feature to set the target of the links in the header and footer to `_blank`. 

Unfortunately we were unable to test the framework as the build process always failed at certain points, but I hope it works this way.

### DIFF
--- a/packages/components/src/components/telekom/app-footer/app-footer.tsx
+++ b/packages/components/src/components/telekom/app-footer/app-footer.tsx
@@ -87,6 +87,7 @@ export class AppFooter {
                         <a
                           class="footer-navigation__item-link"
                           href={item.href || 'javascript:void(0);'}
+                          target={item.target || '_self'}
                           onClick={(event) => {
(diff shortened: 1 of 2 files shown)


niklaswa:
Maybe it would be even better if the attribute is only added to the element if it's not `_self` because it's the default behaviour anyway - but I didn't know what the best syntax for that would be 😅

nowseemee (approved):
Thanks for the PR! 🎉  🎉  🎉

acstll:
> Maybe it would be even better if the attribute is only added to the element if it's not `_self` because it's the default behaviour anyway - but I didn't know what the best syntax for that would be 😅

it could also be `target={item.target}`, if `target` is undefined then the attribute won't render, but that's always not clear (and we've had bugs because of this), so setting it to the default is just right imo.

PR #1593, opened 2023-02-15
fix(nav-icon): add role=button if no href

### LINKED ISSUE #959: Header user menu is implemented as a link and no status is given
Hi,

The profile name is implemented as a link. This gives no clue to a screen reader user that there is a menu opening when clicked and what the state of this menu is (open/closed). Additionally, the horizontal lines are output as „---".

acstll:
Could fix #959

### DIFF
--- a/packages/components/src/components/menu-flyout/menu-flyout.tsx
+++ b/packages/components/src/components/menu-flyout/menu-flyout.tsx
@@ -125,8 +125,13 @@ export class MenuFlyout {
     const triggerSlot = this.hostElement.querySelector(
       '[slot="trigger"]'
     ) as HTMLElement;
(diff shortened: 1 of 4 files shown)


marvinLaubenstein:
Hi @acstll, the PR solves the problem of the issues 🚀. Now it should be understandable enough. Nice solution.

I noticed while checking that JAWS really wants `SPACE` to be the activating instance and we make it accessible via `ENTER`. Is this a problem we should fix ?
⬇️


https://user-images.githubusercontent.com/82942834/219664790-771e098e-4679-4859-811c-148ab5ae958a.mov

acstll:
> I noticed while checking that JAWS really wants `SPACE` to be the activating instance and we make it accessible via `ENTER`. Is this a problem we should fix ?

nice one, let me check, should be possible!

PR #1594, opened 2023-02-16
Fix/tab nav accessibility

marvinLaubenstein:
- Improving tab-panel text accessibility (https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/)
- Fixing problem with clicking on icons within the tab header does not change the tab (also fixes possible accessibility problems due to not selecting after click)

### DIFF
--- a/packages/components/src/components/tab-nav/tab-nav.tsx
+++ b/packages/components/src/components/tab-nav/tab-nav.tsx
@@ -50,20 +50,12 @@ export class TabNav {
     if (svgTags.includes(targetTag)) {
       const closestNextTab = targetHTMLElement.closest(
         `scale-tab-header[role="tab"]`
-      );
-
-      if (!closestNextTab) {
-        return;
-      }
-      this.selectTab(closestNextTab as HTMLScaleTabHeaderElement);
+      ) as HTMLScaleTabHeaderElement;
+      closestNextTab ? this.selectTab(closestNextTab) : null;
+    } else {
+      const nextTab = event.target as HTMLScaleTabHeaderElement;
+      nextTab.getAttribute('role') === 'tab' ? this.selectTab(nextTab) : null;
     }
-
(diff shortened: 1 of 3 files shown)


packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -42,11 +42,20 @@ export class TabNav {
...
+    // workaround for slotted icons
+    const targetHTMLElement = event.target as HTMLElement;
+    const targetTag = targetHTMLElement.tagName.toLowerCase();
+    const svgTags = ['svg', 'g', 'path'];
+
+    if (svgTags.includes(targetTag)) {
+      const closestNextTab = targetHTMLElement.closest(
+        `scale-tab-header[role="tab"]`
+      ) as HTMLScaleTabHeaderElement;
+      closestNextTab ? this.selectTab(closestNextTab) : null;
```
acstll, line 54: This gives an `unused expression` lint error.

packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -42,11 +42,20 @@ export class TabNav {
...
+    } else {
+      const nextTab = event.target as HTMLScaleTabHeaderElement;
+      nextTab.getAttribute('role') === 'tab' ? this.selectTab(nextTab) : null;
     }
-    this.selectTab(nextTab);
```
acstll, line 49: Maybe you can use a `let` for `nextTab` defined before the if/else statement and keep this `this.selectTab(nextTab)` call at the end?

acstll (changes requested):
Except for the lint error, I think this is good to merge. (I make a tiny suggestion only.)

acstll (approved):
💪

acstll:
@marvinLaubenstein LGTM 🙏

Issue #1580, opened 2023-02-08
Segemented Button is not default selected inside Scale Modal


niclas18:
**Scale Version**
3.0.0-beta.122

**Framework and version**
Vue 3.2.47
Vite 4.1.1

**Current Behavior**
When adding a Segmented Button in a Modal, the field marked as `selected` is not selected.

**Expected Behavior**
The field marked as `selected` should also be marked as outside the modal.

**Code Reproduction**
[Stackblitz](https://stackblitz.com/edit/vitejs-vite-fcktqq?file=src%2FApp.vue)

**Desktop:**
- OS: Windows 10
- Browser: Chrome
- Version: 109.0.5414.75

**Additional context**
If you remove the superfluous `v-if` in line 13 of the code reproduction, the component is completely broken.

jotabe87:
I've already commented on the pull request, but for documentation I can also confirm our dev team is facing the same bug. Do you guys have an idea, when the PR will be merged?

felix-ico:
Hi @niclas18 @jotabe87, this issue is not forgotten! I had a look at it and i think it is ok to merge, I will ask another team member to review and hopefully we can get this released this week. 

Just a heads up, I am refactoring some of the segmented button code, but these changes will be included

PR #1597, opened 2023-02-16
fix: no value is set in DropdownSelect and SegmentedButton when using vue-router

niclas18:
Like mentonied in an issue before, the vue-router package make some problems with Scale (#1194).

In our project we have found two new problems:
1. The DropdownSelect have no value set, when setting some default.
2. The SegmentedButton ist not selected it default inside ScaleModal (#1580).

For the first point I have two Sandboxes, one with using vue-router and one without it.
[Sandbox 1: DropdownSelect with using vue-router](https://stackblitz.com/edit/vitejs-vite-dlcfr4?file=src%2FApp.vue)
[Sandbox 1: DropdownSelect without vue-router](https://stackblitz.com/edit/vitejs-vite-atvrmi?file=src%2FApp.vue)

For the second point, there is one Sandbox in my linked issue (#1580).

The behaviour is by both the same. It looks like the vue-router package don't allow using the method `getAttribute()` and `hasAttribute()`, so the return value for it is `null` or `""` (empty string). With the code in this PR the problem is fixed for our project and potentially all users who use the vue-router package or something similiar.

### DIFF
--- a/packages/components/src/components/dropdown-select/dropdown-select.tsx
+++ b/packages/components/src/components/dropdown-select/dropdown-select.tsx
@@ -29,12 +29,15 @@ enum Actions {
   Type = 'Type',
 }
 
+const isElementValue = (x: unknown):x is Element & {value: string} => typeof (x as {value: unknown}).value === 'string'
+const readValue = (element: Element) => isElementValue(element) ? element.value : null
+
 const readOptions = (
   hostElement: HTMLElement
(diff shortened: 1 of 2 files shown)


jotabe87:
Could someone please review and merge this? Our development Team has been having these issues and it seems like this fix is available for quite some time now.

nowseemee (approved):
thank you for the PR!!! 🎉🎉🎉

PR #1610, opened 2023-02-22
fix: add focusable attribute to scale-icons

felix-ico:


packages/components/scripts/icon-component.hbs
```diff
@@ -42,7 +44,7 @@ export class {{className}} {
 
     return (
       <Host>
-        <svg class="scale-icon" xmlns="http://www.w3.org/2000/svg" width={this.size} height={this.size} viewBox="{{viewBox}}" {...ariaHidden}>
+        <svg class="scale-icon" xmlns="http://www.w3.org/2000/svg" width={this.size} height={this.size} viewBox="{{viewBox}}" {...ariaHidden} {...(this.focusable ? { tabindex: 0 } : {})}>
```
acstll, line 47: Wouldn't you put the `this.focusable ? { tabindex: 0 } : {}` in a variable, similar to `ariaHidden` so it's a bit tidier? (this is just aesthetics so don't listen too much)

acstll (approved):

packages/components/scripts/icon-component.hbs
felix-ico: I copied it from scale-icon.tsx 😅 but i could make it cleaner on both files

acstll (approved):
👌

PR #1612, opened 2023-02-22
fix(app-navigation-main-mobile): second level if no children

### LINKED ISSUE #1402: Mobile menu icon label + unnecessary 2nd level of navigation
Scale Version
3.0.0-beta.117

Framework and version
Angular 14

**Current Behavior**
Currently the menu icon label is fixed to the text "close". This should be conffigurable.
Unnecessary 2nd level of navigation in mobile menu even if only one is defined.

**Expected Behavior**
Should be possible to configure a label for the mobile menus' close icon.
There should be no second level of navigation in mobile menu when only on level exists.

**Code Reproduction**
A minimal reproduction case of the bug, you may fork one of the available sandboxes as a base: 

**Desktop (please complete the following information):**

acstll:
Should fix #1402

### DIFF
--- a/packages/components/src/components/telekom/app-navigation-main-mobile/app-navigation-main-mobile.tsx
+++ b/packages/components/src/components/telekom/app-navigation-main-mobile/app-navigation-main-mobile.tsx
@@ -21,6 +21,9 @@ import {
 import { MenuItem } from '../app-interfaces';
 import { findSelected, findRootNode } from '../../../utils/menu-utils';
 
+const hasChildren = (item: any) =>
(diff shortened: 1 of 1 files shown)


acstll:
@nowseemee chef, would you be so kind to take a look at this one? 🪄

nowseemee (approved):
love it!

sulyok-daniel:
I think the issue is still not solved in `3.0.0-beta.126`.

The menu items are correctly recognized having no children - as can be seen at the `aria-haspopup="false"` attribute - in (1), but still we have to do an unneccessary click in (2) to get back to the main menu.

image
image

PR #1628, opened 2023-03-03
Refactor/focus label size

marvinLaubenstein:
- changing focus label size from 10 to 12px 

✅ design review done


felix-ico:
generally this looks good to me, just wondering why `--margin-top-label-focus` is `0.5px` sometimes and `1.5px` sometimes, shouldn't it be the same everywhere? perhaps it's an inconsistency with the box-sizing between the components? 

WDYT @nowseemee @acstll

PR #1631, opened 2023-03-03
fix: remove firstRender attribute and revert to initial autofocus app…

felix-ico:
…roach

this should fix https://github.com/telekom/scale/issues/1372. I am not sure anymore why we did not want an `autoFocus` prop initially and replaced that with a private `firstRender` attribute, but i reverted all changes from [this PR](https://github.com/telekom/scale/pull/1383/files) and added an autofocus prop as the issue reporter suggested

### DIFF
--- a/packages/components/src/components/tab-header/readme.md
+++ b/packages/components/src/components/tab-header/readme.md
@@ -7,13 +7,14 @@
 
 ## Properties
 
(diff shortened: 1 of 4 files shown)


marvinLaubenstein (approved):
Looks good to me 🙂 and definitely gives the issue creator the opportunity to use it correctly.

Issue #1624, opened 2023-02-28
allow icon resize in text-field / helper-text


sulyok-daniel:
For better readability we increased the font size of text-field's error messages.
Unfortunately the icon size is hard coded to 11px - we would like to have a mechanism to configure the icon size, eg.

- by setting it as an attribute on text-field
- by allowing to pass styles to helper-text through text-field
- by binding the icon size to a design token

to match the resized font.

![image](https://user-images.githubusercontent.com/2097315/221816119-2247b9eb-fe77-4bb7-a2c7-a6ee76f17ab2.png)

tshimber:
hello, @felix-ico ! as I see, now **--helper-text-icon-size** variable is available for scale-text-field, and it's great. but it would be good to have the same for scale-dropdown. Now I don't see it, correct me if I'm wrong. 
Would it be possible to add it? Should I create another issue?

Thanks in advance!

felix-ico:
Hi @tshimber, thanks for pointing this out. Of course text-area and dropdown should also have this option. I will take care of this, and there is no need to open a new issue for this. When the PR is ready I will link this issue to it, so you should get notified if you are watching it.

PR #1633, opened 2023-03-03
fix: remove unused css vars, pass helper text vars to helper text com…

felix-ico:
…ponent, update story docs


felix-ico:
this one should fix https://github.com/telekom/scale/issues/1624, or at least make it so that the user can style the helper text color and font size through css vars.

marvinLaubenstein:
Hi @felix-ico, I checked the PR and found that the access via the text-field css variable was not possible. Now it should work. 
I also fixed the attached error (was only the unused constant for the icon size). 
Bildschirm­foto 2023-03-13 um 18 19 43

PR #1640, opened 2023-03-07
feat(header): add slim and subtle types

nowseemee:


packages/components/src/components/telekom/telekom-header/telekom-header.tsx
```diff
@@ -32,6 +32,7 @@ export class TelekomHeader {
   @Prop({ reflect: true }) appName?: string;
   @Prop() appNameLink?: string;
   @Prop() appNameClick?: any;
+  @Prop() type: string = 'default';
```
acstll, line 35: To keep it consistent this would have to be `standard`, no? 🤓

marvinLaubenstein:
⚠️Please merge PR #1669 (https://github.com/telekom/scale/pull/1669) at the same time. Thank you 🙂

packages/components/src/components/telekom/telekom-header/telekom-header.tsx
acstll:
```diff
- default
+ standard
```

it's only 1 character more haha…

But couldn't you make it `''` by default making the `type` prop optional?

nowseemee: yes will do in a follow up PR

PR #1660, opened 2023-03-16
fix(app-navigation-main-mobile): second level if no children (actual child element)

acstll:
Follow up to #1612. Should finally fix https://github.com/telekom/scale/issues/1402 🙏


acstll:
@marvinLaubenstein whenever you get a chance, please check this one out 🙏

marvinLaubenstein (approved):
LGTM🤓, Thanks for finding the solution 🙂

PR #1664, opened 2023-03-17
fix: fix style issue in data-grid pagination

ayeshaa-m:
fix style as per suggestion in the bug.


felix-ico:
Hi @ayeshaa-m, in cases like this we would need to ask one of our designers if it is OK to implement as the user suggests, as this fundamentally just a design change. 

We would also need to update the sketch and figma libraries for consistency, if we merge this.

marvinLaubenstein (approved):

marvinLaubenstein:
LGTM 🤓

Issue #1450, opened 2023-01-06
Scale tag color styling issues after update


goatfryed-holos:
**Scale Version**
3.0.0-beta.117

**Framework and version**
Framework independent

**Previous Behavior**
Previously, it was quite easy to style scale-tag. We just had to provide --background and --color variable, if we had a non interactive simple tag that required custom colors not available in the color palette.

**Current Behavior**
After the update, if the color property is not provided the tag sets additional classes for grey color and displays itself as grey. This overrides most outside color variable configuration, because it has a quite specific selector.

**Expected Behavior**
I'd much appreciate to reintroduce the old behavior. That is, if no color property is set, no additional color classes should apply and the color variables should be set without selectors as defaults.
The current behaviour makes custom styling a real pain

marvinLaubenstein:
Hi @goatfryed-holos, thanks for noticing the problem. Personally, I would also prefer this variant. I'll talk to our team and hopefully change it as soon as possible.

PR #1684, opened 2023-03-21
refactor: adding a possibility to implement user styles

marvinLaubenstein:
Fix: #1450 

=> adding a possibility to implement user styles


felix-ico:
hey @marvinLaubenstein I'm checking this now, I think we should also update the css variables in storybook to reflect these changes

felix-ico:
other than the comment above i tested it and it looks good to me!

PR #1689, opened 2023-03-23
fix: use color token for switch

felix-ico:


felix-ico:
found a bug in the switch in dark mode (for some reason the label color token was defined in storybook but not in the component, so it _looked_ alright in the docs but was actually broken), @marvinLaubenstein @nowseemee could you kindly review

nowseemee (approved):

PR #1690, opened 2023-03-24
fix: reset pagination to last page if new records are rendered are less than previous

ayeshaa-m:
reset pagination start point

### DIFF
--- a/packages/components/src/components/data-grid/data-grid.tsx
+++ b/packages/components/src/components/data-grid/data-grid.tsx
@@ -171,7 +171,7 @@ export class DataGrid {
     this.fieldsHandler();
     this.rowsHandler();
   }
-  componentWillUpdate() {}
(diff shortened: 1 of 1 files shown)


nowseemee (approved):

marvinLaubenstein (approved):

felix-ico:
Hi @ayeshaa-m I'm thinking if we should make this feature optional? I know that some users programmatically add rows to a data grid (for example by populating the data grid with API responses)... currently this implementation would  go back to the first page every time a new row is added to the data-grid. What do you think?

marvinLaubenstein:
@felix-ico Good point. We should make it optional

ayeshaa-m:
@felix-ico @acstll As discussed, the pagination is set to last page if new records are less than the previous records count.

felix-ico (approved):
looks good to me!

PR #1695, opened 2023-03-30
docs(faq): remove code-generated statement

acstll:
Removes a fact in the FAQ section that no longer holds true.


marvinLaubenstein (approved):
LGTM :nerd_face:

Issue #1668, opened 2023-03-17
Notification Toast: Event handlers for closing and closed


georgwittberger-t-systems-mms:
## Description

The Notification Toast component should support event handlers for the following events in its lifecycle:

- User has clicked on the close button or programmatic closing has been initiated (i.e. toast is about to start its fade out animation), e.g. `scale-closing`
- Toast has finished closing (i.e. fade out animation has completed), e.g. `scale-closed`

## Use Case

Application using component libraries like React may have to update internal state when users close the Notification Toast. Considering the fade out animation of the component such state updates must happen after the animation has completed.

## Alternatives

Event handlers can be attached manually to close button element which is complicated and error-prone.

georgwittberger-telekom-mms:
I've just tried the new event `scale-close` in our React project using `@telekom/scale-components-react@3.0.0-beta.131` and it seems to work smoothly. 👍 

Here is how it works in React:

```tsx
<ScaleNotificationToast opened onScale-close={() => { /* do something */ }}>
  <div slot="header">Some Title</div>
</ScaleNotificationToast>
```

marvinLaubenstein:
Really nice, Thanks for your kind message, glad we could solve your problem🙂

PR #1698, opened 2023-03-30
feat: adding close and closing event

marvinLaubenstein:
Feat: #1668


felix-ico (changes requested):
hey @marvinLaubenstein the code seems fine to me, I think we should add the events to the storybook page as well, for consistency

PR #1708, opened 2023-04-05
fix: add scroll container around listbox

ayeshaa-m:
- added a scroll container around list box to resolve blur event call on scroll click.


nowseemee (approved):
thanks for this, @ayeshaa-m 🎉

PR #1714, opened 2023-04-10
fix: fix footer not taking full width of container

ayeshaa-m:
removed max-width property

### DIFF
--- a/packages/components/src/components/telekom/telekom-footer/telekom-footer-content.css
+++ b/packages/components/src/components/telekom/telekom-footer/telekom-footer-content.css
@@ -31,7 +31,6 @@ footer {
 
 [part~='base'] {
   height: 100%;
-  max-width: var(--_max-width-container);
   padding-left: var(--_spacing-x-container);
   padding-right: var(--_spacing-x-container);
   position: relative;


felix-ico:
@ayeshaa-m I think the max-width is actually needed here, in order to make the footer align with the rest of the "grid" (if you check you will find that --scl-grid-max-width is also used in the header, to keep everything the same size)...so I think we need to double check this, maybe setting up a little app-shell and see how it looks when resizing (you can ping me about this for details).

If we end up removing this, I would also remove the private variable `--_max-width-container` since it would not be used anywhere else in the component.

packages/components/src/components/telekom/telekom-footer/telekom-footer-content.css
```diff
@@ -38,6 +41,10 @@ footer {
   display: flex;
   flex: 1;
   flex-direction: column;
+  width: 100%;
```
felix-ico, line 44: I think this would be overwritten by the following width declarations, maybe it could be removed?

felix-ico:
I just tested it out in chrome and ff, chrome looks good but on FF the footer content isn't yet aligned to the header

CHROME: 
Screenshot 2023-04-12 at 11 43 00

FF:
Screenshot 2023-04-12 at 11 43 19

felix-ico (approved):
LGTM

Issue #1713, opened 2023-04-06
Loading Spinner only has English default text for screen reader (Loading)


Tobssen:
**Describe the feature request**
It would be good to have the ability to label the screen reader without showing the label (only for screen reader), so the English text ("Loading") is not shown for all locales.

**Describe the use case**
German Website or different than English.

**Describe alternatives you've considered**
Using the label could be an option, but I do not want it to be displayed.
Customizing Scale is not an option.

**Additional context**
[Add any other context or screenshots about the feature request here.](https://github.com/telekom/scale/blob/main/packages/components/src/components/loading-spinner/loading-spinner.tsx)

PR #1725, opened 2023-04-14
feat: adding accessibility title prop

marvinLaubenstein:
fix: #1713


marvinLaubenstein:
Hi Daniel 🙂, sorry for the late order, but the visual test did not leave me in peace. unfortunately had to find out that the problem also exists in the main branch. 

Could you nevertheless check the PR briefly 🤓?

PR #1731, opened 2023-04-18
refactor: scale button changes for icon large button

ayeshaa-m:
icon large button size is changed to 20px. 
changes are made as per https://www.figma.com/file/9W2ptsoOnpIovWwO0xi84n/Scale-Explorations?node-id=3681-132098&t=1R3lY5pRfOjglp5t-0
https://www.figma.com/file/9W2ptsoOnpIovWwO0xi84n/Scale-Explorations?type=design&node-id=3675-122842&t=vI4UWlZaBzlnDHAj-0


marvinLaubenstein (approved):
LGTM 🚀

acstll (approved):
I think this is perfect, thank you very much @ayeshaa-m 🙌

packages/components/src/components/button/button.css
```diff
@@ -126,8 +126,10 @@
 }
 
 .button.button--icon-before:not(.button--icon-only) ::slotted(*) {
-  margin-right: var(--spacing-icon-x);
-  margin-left: calc(var(--spacing-icon-x) * -1);
+  margin-right: 8px;
```
felix-ico, line 129: i think it would be better here to update the value of --spacing-icon-x instead of hard coding `8px`, unless there is some reason that I am overlooking...

packages/components/src/components/button/button.css
```diff
@@ -137,8 +139,10 @@
 }
 
 .button.button--icon-after:not(.button--icon-only) ::slotted(*) {
-  margin-left: var(--spacing-icon-x);
-  margin-right: calc(var(--spacing-icon-x) * -1);
+  margin-left: 8px;
```
felix-ico, line 142: same as above, since spacing-icon-x changes to 8px everywhere we could set: 

`--spacing-icon-x: var(--telekom-spacing-unit-x2);`

felix-ico (changes requested):
other than the hard coded values everything else looks good to me

felix-ico (approved):
LGTM 👌

PR #1785, opened 2023-05-04
fix: support for prop hides label in iconNavigation and alternate tex…

ayeshaa-m:
support for prop hides label in iconNavigation and alternate text through translation aria labe


nowseemee (approved):
nice, thanks a lot!

could you add usage examples to the fixtures [here](https://github.com/telekom/scale/blob/main/packages/storybook-vue/stories/components/brand-header-navigation/fixtures.js) and [here](https://github.com/telekom/scale/blob/main/packages/storybook-vue/stories/components/beta-brand-header-navigation/fixtures.js), please?

PR #1790, opened 2023-05-04
ARCHIVE: docs: add mobile nav story

felix-ico:


felix-ico:
let's check this again, some mobile navigation docs are needed

PR #1798, opened 2023-05-05
fix: 5 columns behavior on --l --xl

felix-ico:


acstll (approved):
@nowseemee we also need this one shipped 🙏

PR #1800, opened 2023-05-08
fix: header/footer docs

acstll:
- [x] small fix in story for Data API (back compat)
- [x] nicer notification for old deprecated Header/Footer


acstll:
@nowseemee chef, would you be so kind to take a quick look at the code? specially the new `shared.css` file… 🙏 (not the contents but the fact of having the file there…)

nowseemee (approved):
looks good to me!

PR #1822, opened 2023-05-11
fix(radio-button): fix disabled radio button clickable issue

ayeshaa-m:
space between disabled radio button and text should not be clickable

### DIFF
--- a/packages/components/src/components/radio-button/radio-button.tsx
+++ b/packages/components/src/components/radio-button/radio-button.tsx
@@ -81,21 +81,25 @@ export class RadioButton {
   }
 
   handleCheckedChange = (event: any) => {
-    this.checked = event.target.checked;
-    // I don't think this is ever going to be `false` but well...
-    if (this.checked) {
-      this.uncheckSiblings();
+    if (!this.disabled) {
+      this.checked = event.target.checked;
+      // I don't think this is ever going to be `false` but well...
+      if (this.checked) {
+        this.uncheckSiblings();
+      }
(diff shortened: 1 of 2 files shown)


felix-ico (commented):
I tested this and, although the cursor pointer is removed in the space between the button and label (which is good), that area is still clickable and causes the button to be selected

packages/components/src/components/radio-button/radio-button.tsx
```diff
@@ -81,21 +81,25 @@ export class RadioButton {
   }
 
   handleCheckedChange = (event: any) => {
-    this.checked = event.target.checked;
-    // I don't think this is ever going to be `false` but well...
-    if (this.checked) {
-      this.uncheckSiblings();
+    if (!this.disabled) {
```
felix-ico, line 84: i think this one is not really needed but i guess it doesn't really hurt either 🤓

felix-ico (approved):

ayeshaa-m:
> unrelated to this PR but noticed during review: is there a reason why onClick is defined on the whole container and again on the label?

I think we can remove it, but not sure why its added there.

packages/components/src/components/radio-button/radio-button.tsx
ayeshaa-m: I checked the button gets clickable if we do not add this.

felix-ico: oh ok, i thought ` cursor: not-allowed;` would already prevent that on the input field

Issue #1789, opened 2023-05-04
Dropdown Select: Text overflows boundaries of element


georgwittberger-telekom-mms:
**Scale Version**

3.0.0-beta.131

**Framework and version**

React with @telekom/scale-components-react

**Current Behavior**

If text of selected item is very long it overflows the boundaries of the dropdown element.

![dropdown-text-overflow](https://user-images.githubusercontent.com/64543811/236241919-33681241-2400-441a-ba85-35c8e2e0b0a7.png)

**Expected Behavior**

Text of selected item should be capped at the boundary of the dropdown element. Possibly, it should be abbreviated with an ellipsis "..."

**Code Reproduction**

Modify HTML of Scale demo page and insert long text into dropdown item.

**Desktop (please complete the following information):**

 - OS: Windows
 - Browser: Chrome/Edge
 - Version: 112

marvinLaubenstein:
Hi @georgwittberger-telekom-mms thanks for noticing. We will take care of this issue asap.

PR #1825, opened 2023-05-11
Fix/dropdown select 1789

marvinLaubenstein:
Fix: #1789 

Screenshot 2023-05-11 at 18 59 21

### DIFF
--- a/packages/components/src/components/dropdown-select/dropdown-select.tsx
+++ b/packages/components/src/components/dropdown-select/dropdown-select.tsx
@@ -422,7 +422,7 @@ export class DropdownSelect {
               {...(this.helperText ? ariaDescribedByAttr : {})}
               {...(this.invalid ? { 'aria-invalid': 'true' } : {})}
             >
-              {ValueElement}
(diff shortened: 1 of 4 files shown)


packages/components/src/components/dropdown-select/dropdown-select.tsx
```diff
@@ -422,7 +422,7 @@ export class DropdownSelect {
               {...(this.helperText ? ariaDescribedByAttr : {})}
               {...(this.invalid ? { 'aria-invalid': 'true' } : {})}
             >
-              {ValueElement}
+              <span part="combobox_value">{ValueElement}</span>
```
acstll, line 425: I think this should be `combobox-value` with a regular dash (kebab-case) 🧐

acstll (changes requested):
One tiny thing… 🙏

packages/components/src/components/dropdown-select/dropdown-select.tsx
marvinLaubenstein: That definitely makes sense, I will change it. Thank you 🤓

PR #1839, opened 2023-05-16
fix(grid-cell): resolve floating point precision issue

ayeshaa-m:
resolve floating point precision in grid cell

### DIFF
--- a/packages/components/src/components/data-grid/cell-handlers/graph-cell.tsx
+++ b/packages/components/src/components/data-grid/cell-handlers/graph-cell.tsx
@@ -27,7 +27,11 @@ export const GraphCell: Cell = {
     const { style = 'progress', min = 0, max = 100 } = field;
 
     // Convert content to 0>100 range for progress bar
-    const progress = ((content - min) / (max - min)) * 100;
+    const progress = parseFloat(
+      (((content - min) / (max - min)) * 100).toPrecision(
+        String(content).replace('.', '').length
+      )
+    );
 
     switch (style) {
       case 'bar':


packages/components/src/components/data-grid/cell-handlers/graph-cell.tsx
```diff
@@ -27,7 +27,11 @@ export const GraphCell: Cell = {
     const { style = 'progress', min = 0, max = 100 } = field;
 
     // Convert content to 0>100 range for progress bar
-    const progress = ((content - min) / (max - min)) * 100;
+    const progress = parseFloat(
+      (((content - min) / (max - min)) * 100).toPrecision(
+        String(content).replace('.', '').length
```
felix-ico, line 32: I think we should hard code the decimal point value here, or give an optional prop for the user to decide how many decimals should be displayed.

String(content).replace('.', '').length -> this could lead to the same issue again if, for example the content is 19210.9999 it would still render a very long number

ayeshaa-m: Screenshot 2023-05-16 at 2 08 41 PM
I tested with long number as well, using the toPrecision function it is working fine and it will take data as whatever user has provided in datagrid rows.

felix-ico: this was my change to test it:
 
 Screenshot 2023-05-16 at 11 06 34

 and this is the output:
 
 
Screenshot 2023-05-16 at 11 07 04

felix-ico: not sure if i understood something wrong but the i thought the issue was the long decimals in the progress bar precentage

felix-ico: thanks for clarifying @ayeshaa-m, i understood the issue wrong

acstll (approved):
Thank you 🙏

PR #1869, opened 2023-05-23
style(radio-group): added part for fieldset and container

ayeshaa-m:
added part for radio group field set and container

### DIFF
--- a/packages/components/src/components/radio-button-group/radio-button-group.tsx
+++ b/packages/components/src/components/radio-button-group/radio-button-group.tsx
@@ -47,7 +47,7 @@ export class RadioButtonGroup {
   render() {
     const helperTextId = `helper-message-${this.internalId}`;
     return (
-      <fieldset class="radio-button-group">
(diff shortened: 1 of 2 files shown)


packages/components/src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -47,7 +47,7 @@ export class RadioButtonGroup {
   render() {
     const helperTextId = `helper-message-${this.internalId}`;
     return (
-      <fieldset class="radio-button-group">
+      <fieldset part="fieldset" class="radio-button-group">
```
acstll, line 50: `fieldset` is not wrong, but I would name this part `base`. It's probably not evident anymore but we've had a convention to name "base" the root element of the component, in this case:

- `.radio-button-group` → `base`
- `.radio-button-group__container` → `container`

acstll (changes requested):

acstll (approved):

PR #1871, opened 2023-05-23
Fix/a11y chip

marvinLaubenstein:


marvinLaubenstein:
Hi @felix-ico, would be great if you could look over it 🤓. 
Especially about the pre-last commit with the new token for focus outline on dark background.
Thank you very much 🙂

felix-ico:
closing this as chip was approved for a11y & is out of beta

PR #1885, opened 2023-05-25
fix: get selected attribute value for segment

felix-ico:


packages/components/src/components/segmented-button/segmented-button.tsx
```diff
@@ -127,7 +127,7 @@ export class SegmentedButton {
       this.position++;
       tempState.push({
         id: segment.getAttribute('segment-id') || segment.segmentId,
-        selected: segment.hasAttribute('selected') || segment.selected,
+        selected: segment.getAttribute('selected') === 'true',
```
acstll, line 130: Would it work if you just checked for the prop `segment.selected`?

`segment.getAttribute('selected') === 'true'` is flawed, because since `selected` is a boolean attribute, it can have no value (empty string) and still be "true" (plus, the attribute is not `reflect`ed so it can be that `segment.selected` is "true" but the attribute is not there.

I would do both:
- make the `selected` prop reflect (for correctness and convenience)
- then, check only for the prop (not the attribute)

acstll (commented):
Sharing some thoughts ☺️

PR #1899, opened 2023-05-30
fix(date-picker): sync code and figma

nowseemee:
@felix-ico could you take a look at this, please?


felix-ico:
@nowseemee 
Screenshot 2023-05-30 at 16 46 24
Screenshot 2023-05-30 at 16 47 14
i think the top-padding here should be 16px

felix-ico (approved):

PR #1916, opened 2023-05-31
fix: design qa adjustments

felix-ico:


felix-ico:
after further discussion, these changes are not to be implemented

PR #1925, opened 2023-06-01
fix: textarea design alignment fixes

felix-ico:


acstll (commented):
The visual snapshots need a final update 🤓

packages/visual-tests/src/__image_snapshots__/textfield-visual-spec-js-textfield-light-read-only-1-snap.png
acstll, line 1: This is somehow broken but I don't know why 😵‍💫

acstll (changes requested):

acstll (approved):

PR #1932, opened 2023-06-01
fix: ordered list prefix font

felix-ico:


packages/components/src/components/list-item/list-item.css
```diff
@@ -63,12 +63,12 @@
...
   content: attr(data-index) '.';
-  font-size: var(--font-size-marker-ordered);
-  font-weight: var(--font-weight-marker-ordered);
+  font: var(--font-marker-ordered);
 }
 
 .list-item--nested.list-item--ordered:before {
-  font-size: var(--font-size-marker-ordered-nested);
+  font-size: var(--font-marker-ordered-nested);
+  font: var(--telekom-text-style-small-bold);
```
acstll, line 71: I think should be only
```diff
-font-size: var(--font-marker-ordered-nested);
-font: var(--telekom-text-style-small-bold);
+font: var(--font-marker-ordered-nested);
```

acstll (commented):
One tiny thing, the rest in 👌

packages/components/src/components/list-item/list-item.css
felix-ico: you were right, removed

acstll (approved):

PR #1936, opened 2023-06-01
fix: icon size and content spacing, update docs

felix-ico:


packages/components/src/components/collapsible/collapsible.css
```diff
@@ -27,11 +27,11 @@
     var(--telekom-color-functional-focus-standard);
 
   /* icon */
-  --spacing-top-icon: var(--telekom-spacing-unit-x1);
+  --icon-size: 24px;
```
acstll, line 30: I wonder if we want to expose this, since I don't remember having this anywhere else. It's not a bad idea, I'm just not sure.

acstll (approved):

packages/components/src/components/collapsible/collapsible.css
felix-ico: removed it, the icon size was set in the tsx file

Issue #1890, opened 2023-05-26
Scale-Dropdown-Select: hide empty label


Pho3njx:
Dear Scale team,

**Describe the feature request**
As already realized for **_text-field_** and **_dropdown_** (#827 `hideLabelVisually` property) - I would like to ask you to implement this property/feature for the new **_scale-dropdown-select_** too.

**Describe the use case**
When trying to use a **_scale-dropdown-select_** without a label, the text does not fill the empty space above. The space for the label stays empty:
![scale-select-item_noLabel](https://github.com/telekom/scale/assets/95290235/9d96cdd5-db10-433e-8a4a-972716003634)

**Alternatives**
Feel free to realize it in another way.

Thank you and best regards!

ayeshaa-m:
Hi @Pho3njx This feature has been incorporated in upcoming release. Thanks you!

Pho3njx:
Awesome, thank you so much @ayeshaa-m!

PR #1945, opened 2023-06-02
Feat/scale dropdown select- add hide label visibility prop

ayeshaa-m:
add prop to hide label https://github.com/telekom/scale/issues/1890


ayeshaa-m:
Screenshot 2023-06-02 at 2 20 54 PM

packages/components/src/components/dropdown-select/dropdown-select.tsx
```diff
@@ -404,9 +406,11 @@ export class DropdownSelect {
...
         <div part={this.getBasePartMap()}>
           <div part="combobox-container">
-            <label id={`${this.comboboxId}-label`} part="label">
-              {this.label}
-            </label>
+            {!this.hideLabelVisually && (
+              <label id={`${this.comboboxId}-label`} part="label">
+                {this.label}
+              </label>
+            )}
```
acstll, line 413: We shouldn't remove the label completely, it has to remain there for accessibility reasons. So we need to hide it only visually while keeping it in the markup. The old `scale-dropdown` does this, in case it helps.

acstll (changes requested):
Thank you @ayeshaa-m, it looks great, there's only one important detail that requires a small change.

packages/components/src/components/dropdown-select/dropdown-select.tsx
ayeshaa-m: ok Thanks for the input, I will do the change.

acstll (commented):
This is now perfect, but there's a couple of conflicts… 🫣

acstll (approved):

PR #1982, opened 2023-06-07
fix: adjust footer large spacings

felix-ico:


packages/components/src/components/telekom/telekom-footer/telekom-footer.css
```diff
@@ -100,15 +102,18 @@
...
+    column-gap: var(--_nav-items-spacing);
     row-gap: var(--telekom-spacing-unit-x1);
   }
 }
 
 @media screen and (--lg) {
   .scale-telekom-footer :where(ul),
   .scale-telekom-footer[type='minimal'] :where(ul) {
-    column-gap: var(--_nav-items-spacing);
+    --_nav-items-spacing: var(--telekom-spacing-unit-x6);
```
acstll, line 113: Overwriting the var here works, but I think (not 100% sure) the "right" way would be to overwrite it in the original selector where it was defined, so it would affect other places if needed. I mean:

```diff
-.scale-telekom-footer[type='minimal'] :where(ul) {
-  --_nav-items-spacing: var(--telekom-spacing-unit-x6);
+.scale-telekom-footer {
+  --_nav-items-spacing: var(--telekom-spacing-unit-x6);

acstll (approved):

PR #1984, opened 2023-06-07
feat: more visual alignments

acstll:
- pagination
- callout
- progress bar


packages/components/src/components/callout/callout.css
```diff
@@ -6,8 +6,8 @@
   --min-width: 6rem;
   --aspect-ratio: 1;
   --rotation: 0deg;
-  --font-standard: var(--telekom-text-style-body-bold);
   --font-small: var(--telekom-text-style-body-bold);
+  --font-medium: var(--telekom-text-style-body-bold);
```
felix-ico, line 10: is it needed to have these 2 vars when they reference the same token?

felix-ico (approved):
just 1 comment/question but all looks good to me

packages/components/src/components/callout/callout.css
acstll: not really needed but I think it makes sense, at some point one would want to make medium bigger…

PR #2002, opened 2023-06-12
Fix/callout design alignment

felix-ico:


packages/components/src/components/callout/callout.tsx
```diff
@@ -51,8 +51,6 @@ export class Callout {
       characterData: true,
     });
     this.mo = observer;
-    // Wait for full styles before measuring
-    window.requestAnimationFrame(this.adjustSize);
```
acstll, line 55: Let's also remove the `adjustSize` method and the entire MutationObserver stuff… and also make the value of `aspect-ratio` in CSS `1 / 1` instead of `1` (not sure if that really makes a difference, it's only a hunch)

acstll (changes requested):

packages/components/src/components/callout/callout.tsx
felix-ico: removed

acstll (approved):

PR #2018, opened 2023-06-15
fix: add umd out for react wrapper

ayeshaa-m:
add umd out for react wrapper


packages/components-react/.gitignore
```diff
@@ -1,3 +1,5 @@
 node_modules/
 coverage/
 src/components.ts
+lib/
+dist/
```
acstll, line 5: this is probably not needed, `dist/` is already being ignored…

acstll (approved):

PR #2044, opened 2023-06-28
feat(app-header): added interfaces for type safety of navigation props

ayeshaa-m:
I have added interface variables as per code's analysis. Let me know if any field is missing. Kindly check for mandatory and non mandatory fields.


marvinLaubenstein (approved):
LGTM 🚀 (As far as I can judge🤓)

acstll:
More properly typing the `footerNavigation` props somehow breaks it (that's the reason for the visual tests breaking also… the navigation does not show).

PR #2049, opened 2023-07-03
fix: a11y feedback

felix-ico:
- [x] tabfocus not visible on error
- [x] selected list item not output by screen reader
    added a span as discussed
- [x] Error message (helper text) not output by screen reader 

HCM 
- [x] disabled element not visible 
- [x] border not visible 

DISABLED
- [ ] operation instruction is output with wrong pronounciation on disabled drodpown-select
- [ ] element is output with 2 roles: "Problem: Disabled: the screen reader outputs the element with 2 roles= “Ausklappliste” (English: dropdown) and “listbox”. More than one role is confusing for screen reader users."
- [ ] disabled screen reader outputs message twice 

These issues I could not reproduce, there is some label for a disabled dropdown-select referenced in the feedback we got but i could not see it in storybook or anywhere in the code 

Screenshot 2023-07-03 at 17 29 27

### DIFF
--- a/packages/components/src/components/dropdown-select/dropdown-select.css
+++ b/packages/components/src/components/dropdown-select/dropdown-select.css
@@ -285,3 +285,30 @@
 [part~='has-helper-text'] [part~='combobox-container'] {
   margin-bottom: var(--spacing-x-for-helper-text);
 }
+
+.sr-only {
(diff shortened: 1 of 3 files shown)


acstll (approved):
👌

PR #2068, opened 2023-07-21
fix: unnecessary if statement in telekom-nav-item on active watcher

elivlo:
Hi,

recently started working with this library in Vue 3. I created the `scale-telekom-header` and added the `lang-switcher`.

But when I tried to change the `active` prop I ran into the issue that both languages would be "active" (bold font weight).

Here is the template:
```html
<scale-telekom-nav-list variant="lang-switcher" slot="lang-switcher" alignment="right">
   <scale-telekom-nav-item variant="lang-switcher" v-for="lang in $i18n.availableLocales" :key="'mobile'+lang" :active="$i18n.locale === lang">
     <a href="#" @click="changeLocale(lang)">{{ lang.toLocaleUpperCase() }}</a>
   </scale-telekom-nav-item>
</scale-telekom-nav-list>
```

When I entered the code, I found an unnecessary if statement, so I fixed it with this PR.

### DIFF
--- a/packages/components/src/components/telekom/telekom-nav-item/telekom-nav-item.tsx
+++ b/packages/components/src/components/telekom/telekom-nav-item/telekom-nav-item.tsx
@@ -48,12 +48,7 @@ export class TelekomNavItem {
     if (this.linkElement == null) {
       return;
     }
-    if (this.variant === 'lang-switcher' && this.active) {
-      toggleAriaCurrent(this.linkElement, newValue, 'true');
-    }
-    if (this.variant === 'main-nav' && this.active) {
-      toggleAriaCurrent(this.linkElement, newValue, 'true');
-    }
+    toggleAriaCurrent(this.linkElement, newValue, 'true');
   }
 
   connectedCallback() {


elivlo:
> Hi @elivlo and sorry for the wait. I am not sure we should remove that if statement, as the toggleArriaCurrent function is is required for a11y reason.
> 
> The component itself seems to only watch for changes to the active attribute on the component, so it should be up to the consumer to set/remove the active attribute when a nav-item is clicked, this should also apply the correct styling. Would that be doable for you?

Hi @felix-ico,
thanks for your answer.

I am confident that the if statements are bugs. I set the active property, as you can see in my example, but the state only changes if it is set to true. The calls to `toggleAriaCurrent` are never executed when active is set to false.

What do you mean with `consumer`? A vue-slot?

felix-ico (approved):

felix-ico:
>What do you mean with consumer? A vue-slot?

I meant the scale library end-user :)

Anyway I double checked this issue and you are right. I can't seem to merge this PR automatically but i will make sure to include this change for the next release. Thanks again

PR #2069, opened 2023-07-25
fix: wrap element to emit change event

felix-ico:


acstll:
Closing in favor of #2123

Issue #1873, opened 2023-05-23
Dropdown Select: Closes when grabbing scrollbar in listbox


georgwittberger-telekom-mms:
**Scale Version**

3.0.0-beta.131

**Framework and version**

React 18.2.0 with @telekom/scale-components-react

**Current Behavior**

Listbox of the dropdown closes unexpectedly when trying to grab the scrollbar with the mouse pointer.

![chrome_sBRGRja7F9](https://github.com/telekom/scale/assets/64543811/1652098c-1fc8-4377-8ebe-527c2b0b837b)

**Expected Behavior**

Listbox stays open while dragging the scrollbar with the mouse pointer.

**Code Reproduction**

This was not reproducible within the Scale demo page. However, it consistently occurs in our React application using @telekom/scale-components-react.

**Additional context**

My assumption is that the dropdown is closed due to the combobox losing focus, thus executing this `handleBlur` function: https://github.com/telekom/scale/blob/e79a37e81cb54a4803a104b3c0a230c619b32754/packages/components/src/components/dropdown-select/dropdown-select.tsx#L358

For clicks on options there is already a "trick" which sets `this.ignoreBlur = true` to prevent the default closing behavior: https://github.com/telekom/scale/blob/e79a37e81cb54a4803a104b3c0a230c619b32754/packages/components/src/components/dropdown-select/dropdown-select.tsx#L449

Maybe this "ignoreBlur" logic must be added to the `[part="listbox-pad"]` element as well?

marvinLaubenstein:
Hey @georgwittberger-telekom-mms, thank you for noticing. 
Of course, it should not be like that. I'll take a look at it and fix it.

marvinLaubenstein:
Good morning @georgwittberger-telekom-mms, 
I had a look at the issue and indeed it does not occur anymore in the latest version of scale (`3.0.0-beta.134`). I would ask you to update your version. 

If you still have this issue in the updated version, please feel free to contact us again 🙂. 


https://github.com/telekom/scale/assets/82942834/b6a67a15-a054-4c8e-bcf2-48736ea68dbe

georgwittberger-telekom-mms:
@marvinLaubenstein , we've recently upgraded to version `3.0.0-beta.134` and the problem still exists in our application.

[...]

PR #2111, opened 2023-08-29
fix: ignore blur when inside a tab panel

felix-ico:
https://github.com/telekom/scale/issues/1873

### DIFF
--- a/packages/components/src/components/dropdown-select/dropdown-select.tsx
+++ b/packages/components/src/components/dropdown-select/dropdown-select.tsx
@@ -233,7 +233,7 @@ export class DropdownSelect {
   }
 
   connectedCallback() {
-    if (this.hostElement.parentElement.tagName === 'SCALE-TAB-PANEL') {
+    if (this.hostElement.closest("scale-tab-panel")) {
       this.ignoreBlur = true;
     }
     statusNote({ source: this.hostElement, tag: 'beta' });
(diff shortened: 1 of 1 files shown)


packages/components/src/components/dropdown-select/dropdown-select.tsx
```diff
@@ -233,11 +233,15 @@ export class DropdownSelect {
   }
 
   connectedCallback() {
+    if (this.hostElement.parentElement.tagName === 'SCALE-TAB-PANEL') {
```
georgwittberger-telekom-mms, line 236: This would only work if the dropdown is a direct child of the tab panel. However, the issue also occurs when the dropdown is nested somewhere within the tab panel content.

What about this way of checking whether the dropdown is inside a tab panel:

```suggestion
    if (this.hostElement.closest("scale-tab-panel")) {
```

georgwittberger-telekom-mms (changes requested):

packages/components/src/components/dropdown-select/dropdown-select.tsx
felix-ico: I updated the code per your suggestion, but I will still ask about this in the team...honestly i would prefer to remove unnecessary code (if it actually is unnecessary) rather than adding another conditional 😅

PR #2131, opened 2023-09-18
fix(radio-button): polish click handling, hover, etc

acstll:
I removed the `click` handler because it wasn't really needed since the logic was duplicated.

@felix-ico do you see it the same way?

Regarding CSS, `hover` and `active` now target the actual control to avoid triggering them with clicks that trigger no change event… (I hope that makes sense)


felix-ico (approved):
LGTM!

Issue #2119, opened 2023-09-07
href of a tag in brand footer is not configurable und resolves to a bad behavior


tischrei:
**Scale Version**
"@telekom/scale-components": "^3.0.0-beta.134",
"@telekom/scale-components-react": "^3.0.0-beta.134"

**Framework and version**
NextJS
"next": "^13.4.4",

**Current Behavior**
When I inspect the brand footer in my application: there is a href attribut which is set to: `<a href="javascript:void(0);"`
```
<a href="javascript:void(0);" part="logo variant-white transparent" tabindex="0" title="Telekom Logo"><scale-logo-svg part="icon" class="hydrated"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 33 38" part="logo-svg" fill="#ffffff" role="img" aria-labelledby="logo-title-2"><title id="logo-title-2">Telekom Logo</title><path d="M7.6 25.1H0v-7.6h7.6v7.6ZM0 0v12.9h2.3v-.4c0-6.1 3.4-9.9 9.9-9.9h.4V30c0 3.8-1.5 5.3-5.3 5.3H6.1V38h19.8v-2.7h-1.1c-3.8 0-5.3-1.5-5.3-5.3V2.7h.4c6.5 0 9.9 3.8 9.9 9.9v.4h2.3V0H0Zm24.3 25.1h7.6v-7.6h-7.6v7.6Z"></path></svg></scale-logo-svg></a>
```

There is no description in the documentation (https://telekom.github.io/scale/?path=/docs/components-telekom-footer--standard) which specifies an option to set the href to an appropriate value. This leads to the fact, that the SEO statistics are bad on this topic.

![image](https://github.com/telekom/scale/assets/26574398/eb42bbf0-a21c-41d8-94a1-7b33b9b24639)

**Expected Behavior**
An attribute for the brand footer which enables me to change the href attribute to a proper value.

**Code Reproduction**
```
import React from 'react';
import { ScaleTelekomFooter, ScaleTelekomFooterContent } from '@telekom/scale-components-react';
import { defineCustomElements, applyPolyfills } from '@telekom/scale-components/loader';
import Link from 'next/link'

function Footer() {
    return (
        
        <scale-telekom-footer slot="footer">
            <scale-telekom-footer-content>
                <span slot="notice"> © Deutsche Telekom AG </span>
                <ul slot="navigation">
                <li>
                    <Link href="#"> Imprint </Link>
                </li>
                <li>
[...]

PR #2132, opened 2023-09-20
fix: allow setting href on footer logo

felix-ico:
fixes https://github.com/telekom/scale/issues/2119

allow setting footer logo link and title. 

@acstll I'm not really sure how to document these props in storybook though, as they are technically for the child component `telekom-footer-content`, which doesn't have its own storybook page.  Do you have any ideas on that?

### DIFF
--- a/packages/components/src/components/telekom/telekom-footer/telekom-footer-content.tsx
+++ b/packages/components/src/components/telekom/telekom-footer/telekom-footer-content.tsx
@@ -40,7 +40,8 @@ export class TelekomFooterContent {
                 }}
                 transparent
                 href={this.logoHref}
-                title={this.logoHideTitle ? '' : this.logoTitle}
+                logoHideTitle={this.logoHideTitle}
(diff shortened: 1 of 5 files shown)


packages/components/src/components/telekom/telekom-footer/telekom-footer-content.tsx
```diff
@@ -32,6 +39,8 @@ export class TelekomFooterContent {
                     'var(--telekom-line-weight-highlight) solid var(--telekom-color-functional-focus-on-dark-background)',
                 }}
                 transparent
+                href={this.logoHref}
+                title={this.logoHideTitle ? '' : this.logoTitle}
```
acstll, line 43: I think I would this `undefined` instead of an empty string, because undefined (if I'm not mistaken) won't render the attribute, a empty string could render `title=""` (but I'm not sure!)

acstll (approved):
For documenting the prop in the nested component, there's a way. Take a look at the Radio Button Group… 👀 

image

Issue #2130, opened 2023-09-15
app-shell: footer is unreachable when page's heightoverflows only couple of pixels.


susons:
**Scale Version**
    "node_modules/@telekom/scale-components": {
      "version": "3.0.0-beta.137",
      }
    },
    "node_modules/@telekom/scale-components-react": {
      "version": "3.0.0-beta.137",
      }
    },

**Framework and version**
We are using next/js. next: version: 13.1.2.
react: version: 18.2.0.

**Current Behavior**
with the component  imported from 
import { ScaleTelekomHeader } from "@telekom/scale-components-react";

<ScaleTelekomHeader 
      appName={t("")}
      logoHref={''}
      slot="header"
    />
    
  The component on scroll becomes smaller, and if the size diference changes more than the actual overflow, the screen loses overflow and the page gets a jumping page effect and you cannot really react footer in some cases.

**Expected Behavior**
I would expect the component to have attribute scrolled="" only if the actual overflow is more than the difference of the shrinking.

for instance if (overflow > 24px) for mid component screen (size > 1296) only than add the scrolled attribute.

**Code Reproduction**
A minimal reproduction case of the bug, you may fork one of the available sandboxes as a base: 

[Scale and Plain HTML](https://codesandbox.io/s/scale-components-plain-html-template-qfi9mp)
[Scale and React18](https://codesandbox.io/s/scale-components-react18-75ci1x)
[Scale and React18 with Wrapper Package](https://codesandbox.io/s/scale-components-react18-with-wrapper-xqqkrv)
[Scale and Angular13](https://codesandbox.io/s/scale-components-angular13-kxzv2k)
[Scale and Vue3](https://codesandbox.io/s/scale-components-vue3-b14993)

**Desktop (please complete the following information):**
 - OS: windows, Ios
 - Browser [chrome, safari, mozilla]
 - Version [mozilla, 117.0 (64-bit)]

**Additional context**

https://codesandbox.io/s/flyoutmenu-modal-transform-forked-j8v2r9?file=%2Fsrc%2Fstyles.css&resolutionWidth=1632&resolutionHeight=772

susons:
Yes that would be a solution, but not every page has  a lot of content (so a lot of free white space + footer would always have some pixels missing). also this would be rather a big overflow that is always present, because on screen widths larger than 1696px the scroll bar shrink difference is almost 50px, so that all should be then added to the min-height, and in that case there would be always visible scroll bar that is not ideal.

[...]

PR #2133, opened 2023-09-20
fix: set scrolled attribute only when offset is larger than the header

felix-ico:
fix for https://github.com/telekom/scale/issues/2130

### DIFF
--- a/packages/components/src/components/telekom/telekom-header/telekom-header.tsx
+++ b/packages/components/src/components/telekom/telekom-header/telekom-header.tsx
@@ -47,7 +47,7 @@ export class TelekomHeader {
 
   @Listen('scroll', { target: 'document' })
   onScroll() {
-    this.scrolled = window.pageYOffset > 2;
+    this.scrolled = window.pageYOffset > 48;
     this.scrolledBack =
       this.pageYOffset !== window.pageYOffset && window.pageYOffset <= 0;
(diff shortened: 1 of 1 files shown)


packages/components/src/components/telekom/telekom-header/telekom-header.tsx
```diff
@@ -47,7 +47,7 @@ export class TelekomHeader {
 
   @Listen('scroll', { target: 'document' })
   onScroll() {
-    this.scrolled = window.pageYOffset > 2;
+    this.scrolled = window.pageYOffset > 48;
```
acstll, line 50: I would prefer this 48 not be hard-coded, but calculated… if not possible (or too complex), at least add a comment explaining where the number is coming from 🤓

acstll (approved):

packages/components/src/components/telekom/telekom-header/telekom-header.tsx
felix-ico: i thought about calculating it here but it seemed rather overkill for this. Only issue i see with this now, is that on medium viewport the header is not 48px but a little bit smaller - IMO not a problem, it would just minimize when scrolling 10px more than the header height at that viewport size. (or was it requested by UX/UI that the header minimizes ASAP? )

Issue #2109, opened 2023-08-28
Mixup in icons action-arrow-up, action-arrow-down and content-folder


flix-code:
**Scale Version**
3.0.0-beta.136 - but it occurs in the story book examples too!

**Framework and version**
Vue 3.3.4 with Vite 4.4.3

**Current Behavior**
The scale icons `action-arrow-up`, `action-arrow-down` and `content-folder` get mixed up - in the story book examples for all of them the `action-arrow-down` is displayed. Another issue is that if I use the `content-folder`-icon together with the `action-arrow-up`-icon in the same view, not even the same component, the `action-arrow-up`-icon gets replaced by the folder icon.

**Expected Behavior**
If I use an icon I expect it to be exactly that icon and not a mix up or a random other icon.

**Desktop:**
 - OS: MacOS 13.5.1
 - Browser: Chrome
 - Version: 112.0.5615.137 (Official Build) (arm64)

PR #2138, opened 2023-09-21
Fix/optimize icons

felix-ico:
fixes https://github.com/telekom/scale/issues/2109 

(source icons were sharing ids, and for some reason svgo was also updating the ids after i changed them to something more unique, disabling the svgo cleanupIds plugin fixed this)


acstll (approved):
Nice 👌 — did you run [`yarn optimize-icons`](https://github.com/telekom/scale/blob/main/packages/components/package.json#L37) in these src files too?

felix-ico:
> Nice 👌 — did you run [`yarn optimize-icons`](https://github.com/telekom/scale/blob/main/packages/components/package.json#L37) in these src files too?

i did run the optimize-icons script, after disabling the svgo cleanupIds plugin

PR #2159, opened 2023-10-06
fix: allow sorting by german date format

felix-ico:
this fixes a supportpal issue (56456) - sorting by german date format in data grid date cell


acstll (approved):
Lovely 👍

Issue #2162, opened 2023-10-11
Scale-slider range - Right handle covers the left one


Cheesus-Crust1:
**Scale Version**
3.0.0-beta.142

**Current Behavior**
Unable to slide the handle back when reaching the maximum value

**Expected Behavior**
Last used handle should be on top e.g. higher z-index

![slider](https://github.com/telekom/scale/assets/82462082/9f2f3cf5-937d-4938-adb8-bfc048584436)

felix-ico:
this should be fixed in beta.143

PR #2163, opened 2023-10-12
Fix/thumb select

felix-ico:
should fix https://github.com/telekom/scale/issues/2162


acstll (approved):
(Damn failing Switch visual test :neckbeard:)

PR #2175, opened 2023-10-23
fix: add border radius to focussed nav item

felix-ico:


packages/components/src/components/telekom/telekom-nav-item/telekom-nav-item.css
```diff
@@ -132,6 +132,7 @@
 }
 
 .scale-telekom-nav-item > :where(a, button):focus {
+  border-radius: 2px;
```
acstll, line 135: We now have `--telekom-radius-extra-small` for 2px 👀

acstll (approved):
If we can use the token, better…

PR #2185, opened 2023-10-31
feat: add engines to package.json

acstll:
Hopefully saving contributors some time. Using node greater than 16 or not using yarn v1 yields a bunch of install/build errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
     }
   },
   "engines": {
-    "node": ">=16.0.0 <17.0.0",
+    "node": "16.x",
     "npm": "please-use-yarn",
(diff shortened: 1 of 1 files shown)


felix-ico:
tested locally: node v18 throws the correct warnings but i can still use npm without any warnings

acstll:
> tested locally: node v18 throws the correct warnings but i can still use npm without any warnings ( i think the idea is to force yarn right?) ((not sure if something is wrong in my setup))

oh, I think forcing is not needed, maybe that only works with [`engine-strict`](https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict) being true? in any case, it's a bit "documented"

felix-ico (approved):

Issue #2190, opened 2023-11-01
ScaleChip:  [role]s do not have all required [aria-*] attributes


SebastianGode:
**Scale Version**
3.0.0-beta.139

**Framework and version**
NextJS 13.4.4

**Current Behavior**
The Google Lighthouse report throws the following error in the accessibility-test: 
[role]s do not have all required [aria-*] attributes

**Expected Behavior**
The Google Lighthouse report should not throw this error.

**Code Reproduction**
We are using the following Code-Snippet to generate ScaleChips:
```
<ScaleChip
    key={tag.attributes.name}
    onScaleChange={changeChip}>{tag.attributes.name}
</ScaleChip>
```
So this error should exist on all ScaleChips as we didn't specify anything specifically.

**Additional context**
![image](https://github.com/telekom/scale/assets/70581801/ff00c021-d53b-4ad2-9be0-2daab699d048)

PR #2192, opened 2023-11-02
fix: remove unnecessary prop, add required aria attribute

felix-ico:
this should fix https://github.com/telekom/scale/issues/2190

and i removed the optional ariacheckedstate, i guess it can be derived by the selected prop


packages/components/src/components/chip/chip.tsx
```diff
@@ -35,8 +35,8 @@ export class Chip {
   @Prop() selected?: boolean = false;
   /** (optional) chip aria-role */
   @Prop() ariaRoleTitle?: string;
```
acstll, line 37: Instead of the ternary below when defining the `role`,  why not add a default value to the prop? would that work?

```ts
@Prop() ariaRoleTitle?: 'switch' | 'radio' | ... = 'switch';
```

and I would use these as possible types, [according to mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked#associated_roles):
- checkbox
- menuitemcheckbox
- menuitemradio
- option
- radio
- switch

---
btw it's a shame this is called `ariaRoleTitle` instead of `innerAriaRole`, no? 😕

felix-ico: > Instead of the ternary below when defining the role, why not add a default value to the prop? would that work?

this is fixed

> btw it's a shame this is called ariaRoleTitle instead of innerAriaRole, no? 😕

yes but i guess changing this would be a breaking change if people are using the old prop name...

acstll (approved):

PR #2204, opened 2023-11-14
docs: add a vite react example

felix-ico:


acstll (changes requested):
`npm run build` throws for me 😢 

You think it's possible to fix without too much hustle?

image

felix-ico:
@acstll should work now, i updated the dependencies but for some reason yarn still needs "resolutions" to work probperly.

acstll (approved):
👍

PR #1874, opened 2023-05-23
fix tooltip-position inside scaleModal

niclas18:
As described in #1742 there is a problem with the position of tooltips inside modals.
The used package [@floating-ui/dom](https://github.com/floating-ui/floating-ui) have a problem with the shadow-roots there.

They have a fix in there [documentation](https://floating-ui.com/docs/platform#shadow-dom-fix) for this type of problems.
I added this and it fixed the problem for us.

PR #2206, opened 2023-11-15
fix: tooltip within modal

felix-ico:
duplicate of https://github.com/telekom/scale/pull/1874


acstll (changes requested):
Code looks right, but something is making the width funny when text is a few words… (probably the `width: min-content`).

image

(Also, I thought I had removed the Slot story and the `content` slot altogether. A link in there causes a11y problems too. Maybe this is for a different PR…)

felix-ico:
>Code looks right, but something is making the width funny when text is a few words… (probably the width: min-content).

I added a --width css var so that users can set it how they want, in case they have content like in your case, it still defaults to min-content.

packages/components/src/components/tooltip/tooltip.css
```diff
@@ -19,6 +19,7 @@
   --line-height: var(--telekom-typography-line-spacing-standard);
   --spacing: var(--telekom-spacing-composition-space-02)
     var(--telekom-spacing-composition-space-04);
+  --width: 'min-content';
```
acstll, line 22: This is no taking effect because of the quotes. It should be `min-content` not `'min-content'`, no?

If you make it `min-content` it breaks as before. I think we should make this `--width: auto` and then people facing the Modal use-case can set it to `--width: min-content`. Wdyt?

acstll (changes requested):
Almost… 🙂

packages/components/src/components/tooltip/tooltip.css
felix-ico: oh that was my bad, and makes sense to set it to auto 🙏

Issue #2188, opened 2023-11-01
Telekom Brand Header: Logo is focusable without `logoHref` prop


georgwittberger-telekom-mms:
**Scale Version**

3.0.0-beta.142

**Framework and version**

React 18.2.0 with @telekom/scale-components-react

**Current Behavior**

Telekom logo in header is focusable even when no `logoHref` prop is provided. This may be confusing for screen reader users because the generated hyperlink does nothing (`href="javascript:void(0);"`).

**Expected Behavior**

Telekom logo in header is not focusable when no `logoHref` prop is provided. The hyperlink wrapping the logo should have `tabindex="-1"` in this case.

Alternatively, consider not rendering a hyperlink at all if no `logoHref` is given. The logo could be a plain image in this case.

**Code Reproduction**

See Storybook:

- https://telekom.github.io/scale/?path=/docs/components-telekom-brand-header-navigation--standard-header
- https://telekom.github.io/scale/?path=/docs/components-logo--standard

**Additional context**

The issue would have to be addressed in the Scale logo component.

PR #2207, opened 2023-11-16
fix: logo only focusable if href provided

felix-ico:
fixes https://github.com/telekom/scale/issues/2188


acstll (changes requested):
Fixes the issue but then the logo takes 2 tab hits to move focus away from the logo 🤪 — I wonder if it's easy to make it just 1 step. (Not straightforward to figure out what exactly has the focus because of the shadow DOM). Also, probably the `focusable` prop in `scale-logo` could be used?

Plus, this same could be added in `telekom-footer-content.tsx`, no?

acstll (approved):
👌

Issue #2173, opened 2023-10-20
Dropdown Select: Support as form element inside HTML forms


georgwittberger-telekom-mms:
## Description

At the moment, `<scale-dropdown-select>` does not render any HTML element which would make it a form element like `<input>` or `<select>`. Therefore, even when used inside a HTML `<form>` tag and given a `name` attribute, the selected value is not submitted with the form.

For example: Submitting the following form would not transmit the dropdown value:

```html
<form action="/some-action" method="post">
  <scale-dropdown-select label="Select" name="myfield">
    <scale-dropdown-select-item value="opt1">Option 1</scale-dropdown-select-item>
    <scale-dropdown-select-item value="opt2">Option 2</scale-dropdown-select-item>
  </scale-dropdown-select>
</form>
```

Note that the prop `name` is not supported by the dropdown.

As a consequence, `<scale-dropdown-select>` cannot be used in traditional HTML form-based applications without JavaScript workarounds.

Feature request:

1. Provide a prop `name` to allow developers to define a form element name for the dropdown.
2. Ensure that internally some HTML form element (e.g. `input[type=hidden]`) is rendered and updated with the selected value.

## Use Case

We are exploring [Astro](https://astro.build/) as a possible solution for future projects and would like to use the dropdown in conjunction with [HTMX](https://htmx.org/) to submit forms to the server.

## Alternative / Workaround

You can workaround this limitation with some custom JavaScript code. For example using HTMX:

```html
<div>
  <scale-dropdown-select
    label="Select"
    value=""
    data-input-id="dropdown-input"
    hx-on:scale-change="htmx.find(`#${this.dataset.inputId}`).value = event.target.value"
  >
    <scale-dropdown-select-item value="opt1">
      Option 1
    </scale-dropdown-select-item>
    <scale-dropdown-select-item value="opt2">
      Option 2
    </scale-dropdown-select-item>
  </scale-dropdown-select>
  <input id="dropdown-input" type="hidden" name="myfield" value="" />
</div>
```

PR #2212, opened 2023-11-16
fix: add hidden input with name attribute

felix-ico:
this should fix https://github.com/telekom/scale/issues/2173

### DIFF
--- a/packages/components/src/components/dropdown-select/dropdown-select.tsx
+++ b/packages/components/src/components/dropdown-select/dropdown-select.tsx
@@ -211,7 +211,6 @@ export class DropdownSelect {
   /** (optional) is the element used in a form */
   @Prop() formAssociated?: boolean = false;
 
-
   @Event({ eventName: 'scale-change' }) scaleChange!: EventEmitter<void>;
   @Event({ eventName: 'scale-focus' }) scaleFocus!: EventEmitter<void>;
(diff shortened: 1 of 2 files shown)


packages/components/src/components/dropdown-select/dropdown-select.tsx
```diff
@@ -208,6 +208,8 @@ export class DropdownSelect {
   @Prop() ariaLabelSelected?: string = 'selected';
   /** (optional) Text displayed in high contrast mode only to indicate disabled state */
   @Prop() hcmLabelDisabled?: string = 'this field is disabled';
+  /** (optional) is the element used in a form */
+  @Prop() formAssociated?: boolean = false;
```
acstll, line 212: I wouldn't call this `formAssociated` because it's not exactly the same as [the standard static method](https://web.dev/articles/more-capable-form-controls).

I would actually not use any prop at all and just render the hidden input, always. The button has similar "hacks"… What do you think?

acstll (commented):
If you decide to use a prop with a different name, adding it to Storybook would be necessary imo.

packages/components/src/components/dropdown-select/dropdown-select.tsx
felix-ico: makes sense, will do that instead

Issue #1406, opened 2022-12-06
Data-grid 


felix-ico:
We are collecting all issues regarding the data grid here:

https://github.com/telekom/scale/issues/789

https://github.com/telekom/scale/issues/1113

https://github.com/telekom/scale/issues/735

https://github.com/telekom/scale/issues/1199

Issue #1076, opened 2022-05-23
Add color-coding for tags-cell in data-grid


ledex:
**Describe the feature request**
It would be nice to have some sort of color-map for the tags-cell in the data grid so that tags will be colored based on their value.

**Describe the use case**
Being able to see important tags at a glance and have an easier correlation between tags (Currently all tags are grey).

**Describe alternatives you've considered**
N/a

**Additional context**
The definition of the color-map in the tags-field could look something like this:
```javascript
dataGrid.fields = [
  {
    type: 'tags',
    label: 'Status',
    defaultColor: "grey60",
    colorMap: {
      DONE: "green60",
      IN_PROGRESS: "orange60",
      FAILED: "red60"
    }
  }
];
```

Issue #955, opened 2022-03-23
Data Grid Tags Cell


lordichek89:
Hello all,
can i implement such case now? 
![image](https://user-images.githubusercontent.com/4783247/159734191-c0d9887c-c665-41fd-91cc-9f6ecebf32ba.png)

acstll:
Not really, sorry!

(We're planning some updates for the Data Grid in Q2/Q3, so definitely considering this.)

acstll:
#1076 👀

PR #2220, opened 2023-11-24
Fix/data grid tags colors

felix-ico:
from a supportpal request as well as https://github.com/telekom/scale/issues/1406 -> https://github.com/telekom/scale/issues/1076 and https://github.com/telekom/scale/issues/955

this doesn't allow to completely customize the styles of the tags, but at least to use the standard colors we provide https://telekom.github.io/scale/?path=/docs/components-tag--colors#colors

### DIFF
--- a/packages/storybook-vue/stories/components/data-grid/DataGrid.stories.mdx
+++ b/packages/storybook-vue/stories/components/data-grid/DataGrid.stories.mdx
@@ -669,7 +669,7 @@ Expected format: string
 
 ## Tags Cell
 
-Expected format: comma delimited `string`, eg `'one, two, three'`
+Expected format: comma delimited `string`, eg `'one, two, three'` or array of objects with content and color keys, e.g. `{content: 'Apple', color: 'red'}`
 
 <Canvas withSource="close">
   <Story name="Tags Cell">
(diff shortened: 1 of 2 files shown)


packages/storybook-vue/stories/components/data-grid/DataGrid.stories.mdx
```diff
@@ -689,7 +689,7 @@ Expected format: comma delimited `string`, eg `'one, two, three'`
             },
         ];
         dataGrid.rows = [
-            ['Apple, Banana', 'Tomato, Cucumber'],
+            [[{content: 'Apple', color: 'red'}, {content: 'Banana', color: 'yellow'}], 'Tomato, Cucumber'],
```
acstll, line 692: I think this looks awkward… no? 🤔 

image

I would rather document it above:

> Expected format: comma delimited `string`, eg `'one, two, three'` or `array` of objects with `content` and `color` keys, e.g. `{content: 'Apple', color: 'red'}`

What do you think?

felix-ico: @arturo you mean it looks visually awkward and to just document it with some text insted? I did it this way because I thought it would be clear visually and in relation to the code snippet that is shown...

acstll (approved):
🌈

PR #2237, opened 2023-12-08
fix: mega menu grid

felix-ico:
host-context doesn't work in ff and safari...


acstll (approved):
`prettier` check failing… 💅

Issue #2239, opened 2023-12-12
[Tab Navigation]: Programatically enabling tab does not work


ledex:
**Scale Version**
3.0.0-beta.142

**Framework and version**
React 17.0.2

**Current Behavior**
When enabling a tab in the tab navigation programatically (as seen below), the tab button in the navigation becomes enabled again, but navigation to the tab panel does not work.
```JavaScript
<ScaleTabHeader slot={"tab"} disabled={!eventData?.metadata}>
     <ScaleIconAlertInformation/> Metadata
</ScaleTabHeader>
```

**Expected Behavior**
When enabling a tab in the gab navigation programatically it should become available to the user and the user should be able to navigate there.

**Code Reproduction**
https://codesandbox.io/p/sandbox/scale-components-react18-forked-ccr4g4

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox/Chromium

felix-ico:
hi @ledex, this issue should be fixed on the latest release [3.0.0-beta.146](https://www.npmjs.com/package/@telekom/scale-components/v/3.0.0-beta.146)

please feel free to reopen if that's not the case

ledex:
Hi @felix-ico! I cannot reopen this issue, but it seems that this bug is not fixed yet. I've updated my [codesandbox](https://codesandbox.io/p/sandbox/scale-components-react18-forked-ccr4g4) and it still does not work. I can also still reproduce this issue in our code...

felix-ico:
weirdly my initial fix was working in a vanilla js app but not in the react context, this new PR seems to fix the issue completely

ledex:
No worries, I'll try it out once the new version has been released :)

ledex:
Works for me now. Thanks for the fast fix :)

PR #2240, opened 2023-12-18
fix: handle setting disabled programaticcally

felix-ico:
fix for https://github.com/telekom/scale/issues/2239

### DIFF
--- a/packages/components/src/components/tab-nav/tab-nav.tsx
+++ b/packages/components/src/components/tab-nav/tab-nav.tsx
@@ -9,7 +9,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -9,7 +9,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-import { Element, Component, h, Prop, Host, Listen } from '@stencil/core';
```
acstll, line 12: this `Watch` I think it's not needed

acstll (commented):
also `prettier` is missing but other than that is good to go! :shipit:

packages/components/src/components/tab-nav/tab-nav.tsx
felix-ico: ahh fixed

Issue #2081, opened 2023-07-31
autocomplete scale-text-field in dark mode


tk-network:
**Scale Version**
3.0.0-beta.137

**Framework and version**
vue 3.2.45

**Current Behavior**
If the ```scale-text-field``` is filled by autocomplete in dark mode, the labels are no longer perceptible afterwards.

![Bildschirmfoto 2023-07-31 um 10 20 31](https://github.com/telekom/scale/assets/93334942/7e962aee-323f-42d7-9103-edc055476b22)

I have solved the problem like this:
```css
body[data-mode="dark"] input {
    color-scheme: dark;
}
```

But it should still be integrated in the scale package.

**Desktop (please complete the following information):**
 - OS: MAC OS
 - Browser chrome
 - Version 112.0.5615.137

felix-ico:
this bugfix should be available as of beta.139

tk-network:
Hi,

in dark mode it is now correct, but the attribute ```color-schema="dark"``` is now also set in light mode.

![image](https://github.com/telekom/scale/assets/93334942/0f590853-e18b-4a53-8aea-bf21433398da)

![Bildschirmfoto 2023-10-25 um 10 23 55](https://github.com/telekom/scale/assets/93334942/864ec55b-5122-47c0-b54a-e53e241fb8f5)

tshimber:
thanks for this issue, I've already thougth these are some new strange styles//
@tk-network is right, hopefully will be fixed
![image](https://github.com/telekom/scale/assets/110041439/84320283-619b-4fe4-a06e-0a0480613d38)

tk-network:
@felix-ico can you please reopen the ticket?

felix-ico:
this should be fixed in [3.0.0-beta.147](https://www.npmjs.com/package/@telekom/scale-components/v/3.0.0-beta.147)

tshimber:
@felix-ico i've updated to v151, it's still in there:
![image](https://github.com/telekom/scale/assets/110041439/0fa4db62-9ab1-44e5-a57a-2fea461fd1e8)

these are inputs after using chrome autofill

tshimber:
> @tshimber According to the Scale homepage, they are working on a new design system, so I don't think our bug will be fixed. I have solved the problem as follows:
> 
> ```css
> body[data-mode="light"] input {
>     color-scheme: light !important;
> }
> 
> body[data-mode="dark"] input {
>     color-scheme: dark !important;
> }
> ```

[...]

Issue #2245, opened 2024-01-03
Text Field inner colors remain inverted after autofill


pgyoerfi:
**Scale Version**
3.0.0-beta.142

**Framework and version**
Angular CLI: 17.0.8

**Current Behavior**
When autofill suggestions come up inner color go inverted, however, it remaines inverted even if one of the suggestions is accepted.

**Expected Behavior**
Should exit inverted color style when a suggestion is picked by the user.

**Code Reproduction**
A minimal reproduction case of the bug, you may fork one of the available sandboxes as a base: 

[Scale and Plain HTML](https://codesandbox.io/s/scale-components-plain-html-template-qfi9mp)
[Scale and React18](https://codesandbox.io/s/scale-components-react18-75ci1x)
[Scale and React18 with Wrapper Package](https://codesandbox.io/s/scale-components-react18-with-wrapper-xqqkrv)
[Scale and Angular13](https://codesandbox.io/s/scale-components-angular13-kxzv2k)
[Scale and Vue3](https://codesandbox.io/s/scale-components-vue3-b14993)

**Desktop (please complete the following information):**
 - OS: Win11
 - Browser: chrome
 - Version 120
![TextField1](https://github.com/telekom/scale/assets/146071389/b01c8fc9-54e7-46a8-b3ed-d3624dcfe02e)
![TextField2](https://github.com/telekom/scale/assets/146071389/0a1f2b46-e777-4810-9ad7-e670a0fc5c85)
![TextField3](https://github.com/telekom/scale/assets/146071389/41538613-2ecc-444f-98f0-d80969c0356a)

**Additional context**

felix-ico:
this should be fixed in [3.0.0-beta.147](https://www.npmjs.com/package/@telekom/scale-components/v/3.0.0-beta.147), thanks for opening the issue

PR #2247, opened 2024-01-04
fix: prevent chrome default autocomplete colors

felix-ico:
should fix
  https://github.com/telekom/scale/issues/2081
and
 https://github.com/telekom/scale/issues/2245

### DIFF
--- a/packages/components/src/components/text-field/text-field.css
+++ b/packages/components/src/components/text-field/text-field.css
@@ -302,7 +302,8 @@ input:-webkit-autofill,
 input:-webkit-autofill:hover,
 input:-webkit-autofill:focus,
 input:-webkit-autofill:active {
-  color-scheme: dark;
+  /* hack to remove autocomplete colors in chrome, as they conflict with scale dark/light mode styles https://stackoverflow.com/a/68240841 */
+  transition: background-color 0s 600000s, color 0s 600000s;
 }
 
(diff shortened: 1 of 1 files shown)


packages/components/src/components/text-field/text-field.css
```diff
@@ -302,7 +302,8 @@ input:-webkit-autofill,
 input:-webkit-autofill:hover,
 input:-webkit-autofill:focus,
 input:-webkit-autofill:active {
-  color-scheme: dark;
+  /* hack to remove autocomplete colors in chrome, as they conflict with scale dark/light mode styles https://stackoverflow.com/a/68240841 */
```
acstll, line 305: I'm not super happy about this haha… do you think we could try testing with adding `color-scheme: light dark;` globally in :root (in the global CSS file) and remove this `color-scheme` rules for the Text Field only?

felix-ico: hey I did test that already in the old PR for this same issue https://github.com/telekom/scale/pull/2191#discussion_r1384904000 , but just adding `color-scheme: light dark;` to the global css does not fix the issue as far as i can see... (these changes would still be needed for it to work https://github.com/telekom/scale/pull/2191/files)

felix-ico: i mean, it seems like just adding `color-scheme: light dark;` by itself would not fix the issue

Issue #2255, opened 2024-01-11
customizable link cells in data grids


GeorgeDT:
**Describe the feature request**
Currently link cells in a scale data grid are not really customizable. The text represents always the given url without the protocol prefix as it is described in the [docs](https://telekom.github.io/scale/?path=/docs/components-data-grid--link-cell)

**Describe the use case**
Instead of an quite unsightly url containing parameters like Ids etc. an expressive text is shown in the ui.  Clicking the link cell will open the given url / href in the current tab.

**Describe alternatives you've considered**
The data type link cell supports the old unchanged behavior consuming strings. Additional to that the link cell can consume an object with text and props. The text will set the inner text of the scale-link element. The props will pass the given attributes to the scale-link element, in order to set the `href` or the `target` attribute.

PR #2256, opened 2024-01-11
make link cell types customizable

GeorgeDT:
This is intended to address the feature request #2255

The old behavior as well as the new behavior are tested.

E.g. the following two scale data grid columns are defined:

```javascript
    {
      type: 'link',
      label: 'Produkt',
    },
    {
      type: 'link',
      label: 'Edition',
    }
```
And now we can add the rows like this:

```javascript
data.push(
  `https://products.telekom.de/?productId=${product.id}`,
  {text: edition.name, href:`?productId=${product.id}`}
)
```

Which then looks like this:
image

### DIFF
--- a/packages/components/src/components/data-grid/cell-handlers/link-cell.tsx
+++ b/packages/components/src/components/data-grid/cell-handlers/link-cell.tsx
@@ -19,12 +19,24 @@ export const LinkCell: Cell = {
     sortBy: 'text',
   },
   render: ({ content }) => {
-    // Remove protocol (http/https)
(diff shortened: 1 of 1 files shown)


felix-ico:
hi @GeorgeDT, i tested the changes locally and it looks good to me. Could you run `yarn format` and commit the output so that prettier CI job passes?

felix-ico:
Thank you @GeorgeDT!

GeorgeDT:
@felix-ico, thank you very much.

Issue #2251, opened 2024-01-08
Error in icon navigation after scale update


tk-network:
**Scale Version**
3.0.0-beta.146

**Framework and version**
vue 3.3.4

**Current Behavior**
After the update from version Beta 145 to Beta 146 the width is missing in the css class.
In version 145 the width was still set to 100%. As a result, the line below the active menu item is no longer displayed.

**Expected Behavior**
The active menu item should be underlined.

**Code Reproduction**
Beta 145:
![Bildschirmfoto 2024-01-08 um 16 42 50](https://github.com/telekom/scale/assets/93334942/badb85ed-4f18-4bec-9e5b-ee9727940e29)

Beta 146:
![Bildschirmfoto 2024-01-08 um 16 44 07](https://github.com/telekom/scale/assets/93334942/a531df93-7356-402c-ba7a-f73665d4698d)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Chrome
 - Version: 120.0.6099.199

felix-ico:
hi @tk-network this one should be fixed on beta.147 - sorry for the delay, it took me a while to figure out this was happening in `scale-telekom-header-data-back-compat` only 🤓

please feel free to reopen the issue if it's not resolved

PR #2260, opened 2024-01-17
fix(nav-item): hover underline on function slots

felix-ico:
this should fix https://github.com/telekom/scale/issues/2251


tshimber:
@felix-ico maybe you can comment on this PR?
I don't see any visual changes on hover in functions block in header, and in the https://github.com/telekom/scale/issues/2251 you wrote that it's about scale-telekom-header-data-back-compat and is already fixed in version 147. maybe this was some draft PR?

amir-ba:
LGTM , I will merge this

PR #2261, opened 2024-01-19
docs: add angular 17 standalone reactive form sample

felix-ico:


mato-a:
Does this actually work for `scale-dropdown-select[formControl]` or `scale-dropdown-select[formControlName]`?

I get errors like these when I try to use it in Angular 17:

ERROR Error: NG01203: No value accessor for form control unspecified name attribute. Find more at https://angular.io/errors/NG01203

tshimber:
@amir-ba looks good
image

i guess can be merged

amir-ba (approved):
LGTM

PR #2263, opened 2024-01-22
feat(accordion): allow collapsible to be individually expanded

MeridianGH:
A collapsible can now be individually expanded, independent of its containing accordion.

The `expanded` prop is now only propagated, if it is set to true on the parent accordion. This allows the developer to individually select which collapsibles should start expanded instead of all or none.

This change works seamlessly with the `dependent` prop and does not introduce any breaking changes.

### DIFF
--- a/packages/components/src/components/accordion/accordion.tsx
+++ b/packages/components/src/components/accordion/accordion.tsx
@@ -66,7 +66,7 @@ export class Accordion {
     /**
      * Handle `expanded`
      */
-    if (!this.dependent) {
+    if (this.expanded && !this.dependent) {
       this.getCollapsibleChildren().forEach((child) => {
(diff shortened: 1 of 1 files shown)


maomaoZH (approved):

amir-ba:
@MeridianGH great job, can you please rebase it with the main branch so the checks can run for your changes and if they are all green we can merge this. Sorry for the long wait

amir-ba (approved):
LGTM

MeridianGH:
> @MeridianGH great job, can you please rebase it with the main branch so the checks can run for your changes and if they are all green we can merge this. Sorry for the long wait

alright, thank you. the pr should be rebased now, can you check if the pipelines run correctly and if the pr is able to be merged?

PR #2306, opened 2024-03-27
fix(telekom-profile-menu): keyboard accessibility & optional logout handler

tallence-dsulyok:
This PR should fix 2 issues:

1. the profile menu's links & buttons could not be activated via keyboard (since default behaviour was surpressed, even if no clickHandler defined)
2. there was no way to set a custom clickHandler to the profile menu's logout button (even if the underlying component already had the feature that could be used)


felix-ico:
Hi @tallence-dsulyok the code looks fine to me, i also tested locally. One minor thing is, maybe we should add the new logoutHandler in the storybook docs, so people know this prop exists as well?

felix-ico:
@tallence-dsulyok there is also a ci job failing (build-pr / uncommitted-changes), could you run `yarn build` in your components folder and commit the changes?

PR #2309, opened 2024-04-04
fix(switch): add text color for switch toggle

jwueller:
Introduces a `--_on-background` variable that's used to define the IO text color relative to the `--_background` color of the toggle, and applies it to the base state.

## The Problem

Black background with white text; both the label and IO text have poor to no contrast.
grafik

This can be mitigated by setting `--color-label` to `currentColor` to honor the chosen text color. Maybe this should even be the default, considering the label is always going to be on top of the surrounding background? The contrast problem remains for the IO text, and is not adjustable by the user.
grafik

## The Solution

Since the switch toggle itself already has a background color, contrast becomes independent of the surrounding background. It's therefore inconsistent to honor `--color-label` in this context. The new variable ensures that the base states color actually related to the background it's drawn on.
grafik

It currently defaults to `var(--telekom-color-text-and-icon-standard)`, which also behaves correctly when the control is used in dark mode.
grafik

### DIFF
--- a/packages/components/src/components/switch/switch.css
+++ b/packages/components/src/components/switch/switch.css
@@ -26,6 +26,7 @@ scale-switch {
 
 .switch {
   --_background: var(--telekom-color-ui-faint);
+  --_on-background: var(--telekom-color-text-and-icon-standard);
   --_color-thumb: var(--telekom-color-ui-white, #fff);
   --_overlay-background: transparent;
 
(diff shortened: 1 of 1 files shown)


maomaoZH (approved):
@jwueller  Thanks for the PR. LGTM!

tshimber:
@amir-ba i guess can be merged

amir-ba (approved):
@jwueller thanks for this, and sorry for the late response. Can you please merge the main branch into this, so the checks can run correctly again. We will merge this afterward

Issue #2282, opened 2024-02-20
[Tab Navigation] Programmatically select tabs


ledex:
**Describe the feature request**
This feature would allow the current tab to be programmatically selected based on the state of the application.

**Describe the use case**
This feature has many use cases and has been implemented by other UI-libaries.   

In the example below, many tabs are disabled since the event is dropped and only the selection filter can be displayed. Without this feature, the first tab is always selected, even if it's disabled.
![grafik](https://github.com/telekom/scale/assets/39191120/5a020aab-a9e6-4e66-adf9-62ebcba357ee)

**Describe alternatives you've considered**
Our current workaround is that we have a ref to the "Selection Filter" tab (from the example above) and we conditionally click on the ref.
```javascript
if (event.status === "DROPPED") {
    sfTab.current.click();
}
```

**Additional context**
MUI i.e. allows to set the index of the initially selected tab as documented [here](https://mui.com/material-ui/api/tabs/#tabs-prop-value).

tshimber:
@ledex hi! where's no info, which version of Scale do you use, would be helpful. anyway please update to latest version. in this case:

1) disabled tabs shouldn't got selected. I've checked it in code and tested it: the first enabled tab is getting selected on start.

2) considering concrete tab to select: this feature is supported as well. to use it, you need to put 'selected' on the **tab header**, which should be selected on start. this is not obvious, so we meight add it to the story book. thus adding additional tab-index-to-select prop seems excessing.

please, leave a feedback if it works for you.

tshimber:
example1: first tab is disabled
<scale-tab-nav>
      <scale-tab-header disabled slot="tab">General</scale-tab-header>
      <scale-tab-panel slot="panel">
        Freegan kinfolk farm-to-table humblebrag cred…
      </scale-tab-panel>
      <scale-tab-header slot="tab">Usage</scale-tab-header>
      <scale-tab-panel slot="panel">
        Bespoke austin pork belly yuccie pop-up. Before they sold out…
      </scale-tab-panel>
      <scale-tab-header slot="tab">Style</scale-tab-header>
      <scale-tab-panel slot="panel">
        Biodiesel chia af hoodie tumeric bespoke letterpress…
      </scale-tab-panel>
      <scale-tab-header slot="tab">Code</scale-tab-header>
      <scale-tab-panel slot="panel">
        Asymmetrical tattooed chia, banh mi blog microdosing…
      </scale-tab-panel>
    </scale-tab-nav>
    
    
[...]

PR #2320, opened 2024-06-11
fix(tab-nav): fixes preselect logic in tabs with disabled attribute

tshimber:
Fixes https://github.com/telekom/scale/issues/2282

- now if selected tab gots disabled, selection moves to first available tab
- example of preselected tab logic added to storybook

### DIFF
--- a/packages/components/src/components/tab-header/tab-header.tsx
+++ b/packages/components/src/components/tab-header/tab-header.tsx
@@ -47,7 +47,7 @@ export class TabHeader {
   /** (optional) size  */
   @Prop() size?: 'small' | 'large' = 'small';
   /** (optional) Whether the tab is selected */
-  @Prop() selected?: boolean;
+  @Prop({ mutable: true }) selected?: boolean;
   /** (optional) Injected CSS styles */
   @Prop() styles?: string;
 
@@ -64,7 +64,7 @@ export class TabHeader {
     if (this.disabled) {
       return;
     }
-    this.scaleSelect.emit();
+    this.selected = true;
   }
 
   @Watch('selected')
@@ -73,6 +73,9 @@ export class TabHeader {
       return;
     }
     if (!this.disabled) {
+      if (newValue === true) {
+        this.scaleSelect.emit();
+      }
       if (newValue === true && this.tabsHaveFocus()) {
         // Having focus on the host element, and not on inner elements,
         // is required because screen readers.
--- a/packages/components/src/components/tab-header/tab-header.tsx
+++ b/packages/components/src/components/tab-header/tab-header.tsx
@@ -65,6 +65,7 @@ export class TabHeader {
       return;
     }
     this.selected = true;
+    this.scaleSelect.emit();
   }
 
   @Watch('selected')
--- a/packages/components/src/components/tab-header/tab-header.tsx
+++ b/packages/components/src/components/tab-header/tab-header.tsx
@@ -56,7 +56,7 @@ export class TabHeader {
   /** Emitted on header select */
   @Event({ eventName: 'scale-select' }) scaleSelect: EventEmitter;
   /** Emitted when currently selected tab got disabled */
-  @Event({ eventName: 'scale-got-disabled' }) scaleGotDisabled: EventEmitter;
+  @Event({ eventName: 'scale-disabled' }) scaleDisabled: EventEmitter;
 
   @Listen('click')
   handleClick(event: MouseEvent) {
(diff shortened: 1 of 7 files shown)


maomaoZH (approved):
@tshimber  thanks for the PR. LGTM.

tshimber:
> @tshimber thanks for the PR. LGTM.

@maomaoZH thank you! but unfortunately it doesn't work in deployed version🫠 I try to understand now, how these storybook templates are working. sorry, i'll put the draft mark now..

maomaoZH:
@tshimber  Scale is still using storybook 7.  So it is a bit tricky to add story.

packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -171,10 +185,24 @@ export class TabNav {
   }
 
   selectTab(nextTab: HTMLScaleTabHeaderElement) {
+    this.resetTabs();
+    nextTab.selected = true;
+    this.selectPanel(nextTab);
+  }
+
+  selectPanel(nextTab: HTMLScaleTabHeaderElement) {
```
amir-ba, line 193: Separating the panel selection from the `tabselection` makes the sync harder. I suggest having one single function that selects panel and tab together

packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -171,10 +185,24 @@ export class TabNav {
...
+    this.resetPanels();
     const nextPanel = this.findPanelForTab(nextTab);
-    this.reset();
     nextPanel.hidden = false;
-    nextTab.selected = true;
+  }
+
+  deselectTabs(nextTab: HTMLScaleTabHeaderElement) {
```
amir-ba, line 199: make the current resetTab a general function that can handle this
```suggestion
const tabs = this.getAllEnabledTabs();
    tabs.forEach((tab) => {
      if (!nextTab || tab !== nextTab) {
        tab.selected = false;
      }
    });
```

packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -42,9 +42,16 @@ export class TabNav {
   handleSelect(event) {
     const nextTab = event.target as HTMLScaleTabHeaderElement;
     // Act only if it's a direct child
-    if (this.getAllEnabledTabs().includes(nextTab) && !nextTab.disabled) {
-      this.selectTab(nextTab);
+    if (!this.getAllEnabledTabs().includes(nextTab) || nextTab.disabled) {
```
amir-ba, line 45: This way, the added return call will be useless. I suggest you return this to its previous form and adapt the `selectTab`

packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -171,10 +185,24 @@ export class TabNav {
   }
 
   selectTab(nextTab: HTMLScaleTabHeaderElement) {
+    this.resetTabs();
+    nextTab.selected = true;
+    this.selectPanel(nextTab);
+  }
```
amir-ba, line 191:
```suggestion
  selectTab(nextTab: HTMLScaleTabHeaderElement) {
    if (!nextTab.selected) {
      nextTab.selected = true;
      const nextPanel = this.findPanelForTab(nextTab);
      nextPanel.hidden = false;
    }
    this.resetTabs(nextTab);
  }
```

packages/components/src/components/tab-nav/tab-nav.tsx
```diff
@@ -146,22 +150,30 @@ export class TabNav {
 
   linkPanels() {
     const tabs = this.getAllTabs();
-    const selectedTab =
-      tabs.find((x) => x.selected) || tabs.filter((x) => !x.disabled)[0];
-
```
tshimber, line 151: i'm still thinking this functionalyty must be in separate method, it took me some time to find it here when i started

packages/storybook-vue/stories/components/tab-navigation/ScaleTabNav.vue
```diff
@@ -55,6 +55,7 @@ export default {
   props: {
     styles: { type: String },
     disabled: { type: Boolean, default: false },
+    preselected: { type: Boolean, default: false },
```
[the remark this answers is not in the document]
tshimber: without it we can't have these upper visualization, and visualization is nice\\ i guess it's not the first time some extras are added only in storybook to visualize stuff, i'm pretty sure i've made it by analogy 
image

packages/components/src/components/tab-header/tab-header.tsx
```diff
@@ -47,42 +47,46 @@ export class TabHeader {
...
+  @Event({ eventName: 'scale-disabled' }) scaleDisabled: EventEmitter;
 
   @Listen('click')
   handleClick(event: MouseEvent) {
     event.stopPropagation();
     if (this.disabled) {
       return;
     }
-    this.scaleSelect.emit();
+    this.selected = true;
```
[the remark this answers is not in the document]
tshimber: at first there was a loop because of it, but then i've added some checks on parent, now it's ok, so you're right, better bring emitter back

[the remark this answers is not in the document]
tshimber: now emitter is back

packages/components/src/components/tab-nav/tab-nav.tsx
tshimber: i'm not sure what you mean, i put this stuff back into linkPanels as separate call and renamed linkPanels to   linkPanelsAndSelectTab. or how you see it?

packages/storybook-vue/stories/components/tab-navigation/ScaleTabNav.vue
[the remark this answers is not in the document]
tshimber: what do you think? or we can discuss in call.

packages/components/src/components/tab-nav/tab-nav.tsx
amir-ba: you can make this a separate function because this also gets called in other places. and they can both reference this function 
```js
const selectTab = (tabs) =>  tabs.find((x) => x.selected) || tabs.filter((x) => !x.disabled)[0];
```

tshimber: now understood, hehe

packages/storybook-vue/stories/components/tab-navigation/ScaleTabNav.vue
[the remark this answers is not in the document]
tshimber: it appears here automatically\\ let's discuss tomorrow

Issue #2327, opened 2024-07-11
Accessibility | Dropdown component - no automatic scrolling for keyboard users


kolianovich:
**Scale Version**
"@telekom/scale-components": "3.0.0-beta.152",

**Framework and version**
"react": "^18.2.0" - without wrapper

**Current Behavior**
When an user with keyboard navigates through opened options list - there is no scrolling to selected element and focus goes outside of visible area.

![image](https://github.com/telekom/scale/assets/25793946/df21d17b-ef3e-46f8-b035-1793f14ce5af)

**Expected Behavior**
When an user with keyboard navigates through opened options list - focused element should always be inside visible area. It should be achieved by auto scrolling of option list to focused element.

**Additional context**
We kindly request to fix it asap because of this issue has 3 thumbs down in accessibility report what considered as critical accessibility issue.

PR #2333, opened 2024-07-31
dropdown scroll fixed

tshimber:
Fixes https://github.com/telekom/scale/issues/2327
Scrolling issue if operating with keyboard for dropdown-select component fixed. Seems to be the bug from previous refactor (element ref on wrong container)


amir-ba:
LGTM. This seemed to be missed from latest refactors. Thanks for this.

Issue #2326, opened 2024-07-09
Accessibility | Data Grid Component settings menu requires tableOptions localization


KonstantinBenzing:
**Scale Version**
3.0.0-beta.151

**Framework and version**
Basic components

**Current Behavior**
The settings menu icon has a hard-coded accessibilityTitle "Table Options", which is in English. A screen reader will always read English, even though the person might not understand English.

**Expected Behavior**
By means of using the localization object, I expected to be able to change the accessibility title to another language.

**Code Reproduction**
  @Prop() localization?: {
    sortBy: string;
    toggle: string;
    select: string;
    **tableOptions: string;**
  };
...

<scale-menu-flyout class={`${name}__settings-menu`}>
        <scale-button
          slot="trigger"
          variant="secondary"
          icon-only
          data-sortable={this.isSortable}
        >
          <scale-icon-service-settings
            accessibilityTitle={this.localization?.tableOptions || "Table Options"}
          ></scale-icon-service-settings>
        </scale-button>
...

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 126.0.6478.126 (Official Build) (64-bit)

**Smartphone (please complete the following information):**
n/a

**Additional context**
For accessibility the localization object should allow to change all hard-coded text fragments, which could potentially be read by a screen reader. This counts as level 2 accessibility issue in accessibility reports and can cause denial of acceptance.

PR #2334, opened 2024-07-31
table options localization added

tshimber:
Fixes https://github.com/telekom/scale/issues/2326
Localization obj for data-grid component exended with tableOptions field


amir-ba (approved):

amir-ba (commented):
Please add this to the docs

Issue #2202, opened 2023-11-09
defaultSorting of scale-data-grid


HejHG:
**Describe the feature request**
I want to define the default sorting Column inside scale-data-grid including shown sorting arrow on header of the column

**Describe the use case**
yes, I can sort the data shown in the grid before hand over to the dataGrid, but I also want that the user can see the sorting on the header of the column.

**Describe alternatives you've considered**
sorting data on my own, but then I can not set the sorting arrow on the column header

**Additional context**
non

PR #2335, opened 2024-08-08
feat(data-grid): adds presort to grid

tshimber:
Fixes https://github.com/telekom/scale/issues/2202

### DIFF
--- a/packages/components/src/components/data-grid/data-grid.tsx
+++ b/packages/components/src/components/data-grid/data-grid.tsx
@@ -183,7 +183,6 @@ export class DataGrid {
     this.fieldsHandler();
     this.rowsHandler();
   }
-  componentWillUpdate() { }
   componentDidRender() {
     if (this.needsAutoWidthParse) {
       this.calculateAutoWidths();
@@ -198,7 +197,7 @@ export class DataGrid {
   componentDidLoad() {
     this.addResizeObserver();
   }
-  componentDidUpdate() { }
+
   disconnectedCallback() {
     this.removeResizeObserver();
   }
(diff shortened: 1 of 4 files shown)


packages/components/src/components/data-grid/data-grid.tsx
```diff
@@ -493,6 +503,21 @@ export class DataGrid {
     this.activeSortingIndex = -1;
   }
 
+  presortIfNeeded(): void {
```
amir-ba, line 506: Better call the fn `preSort` the if needed part is implied

packages/components/src/components/data-grid/data-grid.tsx
```diff
@@ -416,15 +420,21 @@ export class DataGrid {
...
         ? 'ascending'
-        : sortDirection === 'ascending'
+        : currentSortDirection === 'ascending'
         ? 'descending'
         : 'none';
     this.fields[columnIndex].sortDirection = newSortDirection;
     this.sortTable(newSortDirection, type, columnIndex);
   }
 
+  presortTable(sortDirection, columnIndex, type): void {
```
amir-ba, line 432: Why not send the `column`  itself?  Since you are finding an index once in the presort function and finding the field here again. We can skip that part

packages/components/src/components/data-grid/data-grid.tsx
tshimber: renamed and merged methods

packages/components/src/components/data-grid/data-grid.tsx
tshimber: this logic is used initially in template and other sorting methods, changing causes changes across the whole file, but I can)

amir-ba:
We still would need the storybook extension.

amir-ba:
a storybook example for this would be nice, and you might need to merge the main branch in the branch to fix the visual tests

tshimber:
presort example added to doc, visual tests are fine

Issue #2331, opened 2024-07-29
telekom-profile-menu


EdsonT-Systems:
**Scale Version**
Latest

**Framework and version**
Angular latest with   auto-generated scale wrapper packages

**Current Behavior**
A clear description of what the bug is and how it manifests.

- If I try not to use properties like serviceLinks(because for this app I don't need them), the component doesn't render.

- I can't find a way to correctly apply the logoutHandler method to implement the logout logic because it doesn't recognize that function.

**Expected Behavior**

- I am trying to apply a logout logic with the previously mentioned method, but the component doesn't allow me to.

**Code Reproduction**

**Desktop (please complete the following information):**
 - OS: [windows]
 - Browser [ chrome,]
 - Version [latest]

tshimber:
hi @EdsonT-Systems !
1) I've checked: no serviceLinks should work fine, but it fails without userInfo. I've pushed a fix, we will deploy new release in a while. If it fails for you with userInfo, please provide usage example, I'll check again

If you wanna provide userInfo in Angular:
template:
`[userInfo]="userInfo"`
ts: 
`get userInfo() {
    return {
      name: 'Alexander Dreyer',
      email: 'alexander.dreyer@t-online.de',
    };
  }`

2) also provide the example how you use logoutHandler, please. For me it works fine in Angular: 
template:
`[logoutHandler]="handlerFunc"`
ts: 
`handlerFunc(): void { console.log('LOGOUT'); }`

note that if you provide logoutHandler, logoutUrl won't work -- these are alternative options

EAristiguieta:
Hi @tshimber 

Well, after trying many options, I decided to use personalize component because the option you mentioned using the function works in the case of the console.log, but the navigation I am trying to do doesn't work properly since it doesn't render the rest of the components correctly. It's hard to explain with images, but basically, I tried the navigation using that function in that component, and it didn't work. I did it with custom components, and it worked fine.

Example:

1-Using the [logoutHandler]="handlerFunc"

-Ts:
![image](https://github.com/user-attachments/assets/5c9cf015-23e5-4275-bc2e-722037695d08)

-Html:
![image](https://github.com/user-attachments/assets/c4ae1271-dcd1-42ad-8dc5-7093d659b1dc)

-Ui:

![image](https://github.com/user-attachments/assets/5847b474-9a5c-49da-9e5f-d4d513e401f4)


-At this moment when the log out its clicked 
[...]

Issue #2293, opened 2024-03-11
Telekom header language switcher not visible


nikolab-pn:
**Scale Version**
beta 150

**Framework and version**
React

**Current Behavior**
Language switcher buttons are moved outside of viewport

**Expected Behavior**
Buttons should be visible

**Code Reproduction**
.scale-telekom-nav-item > button in telekom-profile-menu.css should be scoped so it doen't change language switcher button padding

tshimber:
hi @nikolab-pn ! 
could you please clearify, what component is affected as there are deprecated and new headers. and maybe add screenshot?
in the new header lang swither seem to look ok for me:
![image](https://github.com/user-attachments/assets/c1ee73aa-a016-4a8b-a998-82e1ff6f3434)
![image](https://github.com/user-attachments/assets/a693da80-e86e-4a30-ad5b-23d3d122aa64)

 -- it's here also with profile menu, it disappears on scroll

nikolab-pn:
The issue is visible if inside `scale-telekom-nav-item` instead of `a` element `button` element is used. It's not a big issue, but it's also easy fix

How it looks
<img width="167" alt="image" src="https://github.com/user-attachments/assets/cb234e1b-2bae-4dd4-815b-65c4526f0c7e">


How it should look
<img width="170" alt="image" src="https://github.com/user-attachments/assets/138aceeb-d9a4-42e2-8524-96fac3327cc8">

tshimber:
@nikolab-pn thank you for clarification! we'll fix it

PR #2339, opened 2024-08-21
no userInfo for profile-menu 

tshimber:
Fixes https://github.com/telekom/scale/issues/2331
Fixes https://github.com/telekom/scale/issues/2293
telekom-profile-menu component was failing in case userInfo is not provided. The check for empty prop is added.
UPD: no serviceLinks + hideLoginSettings case fixed.

UPD: lil css nesting fix for nav item inside profile menu added

@amir-ba we could discuss how css should be refactored for components having shadow: false. now looks like muddle which i'm afraid to touch


amir-ba:
LGTM!

amir-ba (approved):

PR #2342, opened 2024-08-27
ci: sets up docker compose alias for compatibility

amir-ba:
This PR attempts to fix the failing visual-test job in the workflow.  It first removes the unneccessary `ref` for the `actions/checkout@v3` as it tries to pull references that might not exist in addition based on the latest [Docker changes ](https://docs.docker.com/compose/migrate/) the `docker-compose` used by our `docker-chromium` dependency is removed from July 2023, and we need to provide a Symlink that calls the new `docker compose` v2.


acstll (approved):
Seems to work! 🙂

Issue #2315, opened 2024-05-23
[ScaleNotification] Events in Angular


ch4d1:
**Scale Version**
"@telekom/scale-components": "^3.0.0-beta.151",

**Framework and version**
Angular 17.3.0

**Code Reproduction**
[Scale and Angular17](https://codesandbox.io/p/devbox/scale-components-angular17-ws7g2p)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Google Chrome
 - Version: 125.0.6422.77

```
<div class="scl-toast-stack">
  <scale-notification
  heading="Warning toast headline"
  type="toast"
  variant="warning"
  dismissible
  opened
  (scaleOpen)="test('open')"
  (scaleBeforeClose)="test('BeforeClose')"
  (scaleClose)="test('Close')"
  >
    <span slot="text">Lorem ipsum dolor sit amet, consetetur tempordede invidunt ut labore et dolore.</span>
  </scale-notification>
</div>
```

**Additional context**
It is very likely a user issue, as I am not an expert in using Angular. However, I need help with the usage of events from scale components. I am currently trying to capture events for a scale-notification, but these events do not seem to exist. I have created an example with CodeSandbox. I would expect that the events when closing a notification would trigger the function "demo". However, nothing happens. Am I missing something?

ch4d1:
Thank you for the clarification! I switched to using the kebab-case notation for the event, and everything is working now. I can’t believe I didn’t try that sooner. However, IntelliJ IDE is still flagging the Event by saying, "Event scale-open is not emitted by any applicable directives nor by the <scale-notification> element."

It might be something specific to my setup, but I wanted to double-check. I do know that the React implementation doesn’t have this issue when using the proxy package... but for Angular, it's deprecated and outdated.

Thanks again for the help!

tshimber:
Hi @ch4d1 !
Strange about the IDE, I don't see such in VS code, and there's an emit for scale-open on 'opened' prop changing:
![image](https://github.com/user-attachments/assets/f0ada7cb-0bd5-442c-b686-6ae80b393496)
![image](https://github.com/user-attachments/assets/dfcdef26-e477-41cf-aee6-ca9af14af73f)

and it fires:
![image](https://github.com/user-attachments/assets/a68e56e7-a7d0-4dcb-96fa-ec699eea81c1)
![image](https://github.com/user-attachments/assets/b4e46a1d-b194-465f-baf1-31d67ae779d8)

[...]

PR #2345, opened 2024-08-30
event part in storybook fixed

tshimber:
Fixes https://github.com/telekom/scale/issues/2315
camelCase to kebab-case event naming info added

PS I've checked, it works as kebab-case for Angular and Vue, also for React both with .addEventListener and with wrappers.


amir-ba:
LGTM!

amir-ba (commented):
LGTM

PR #2351, opened 2024-09-11
Fix: submit button does not add name or value to submit form event

marcrobertscamao:
The name and value used on the button element are not submitted as part of the form event as these are not cloned onto the temporary button used for submitting the form.


amir-ba:
LGTM, I will merge this. thank you

Issue #2349, opened 2024-09-11
Usage of "aria-detailed-id"


EdsonT-Systems:
**Scale Version**
Latest

**Framework and version**
Angular latest with   auto-generated scale wrapper packages

**Current Behavior**
The "aria-details-id" property doesn’t work without the "helper-text" property.

**Expected Behavior** **Code Reproduction**
The expected behavior is for the property to work without relying on others.

In the project I’m working on, we use the error message in a different element without using the "helper-text". The issue is that when using "aria-detailed-id" (to link the error message with the input field) without the "helper-text", the NVDA screen reader reads the property but says it’s empty.

Is there any way to use "aria-detailed-id" and have it recognized without using the "helper-text" property?

Example:

![image](https://github.com/user-attachments/assets/5b221ee7-d72b-4ad9-a5ff-5868b9f3506d)

@tshimber @acstll @felix-ico @amir-ba

tshimber:
hi @EdsonT-Systems ! i've checked the code, this attribute doesn't seem to have any connection to helperText. if value is provided, it should convert into aria-details attribute on <input>, and it works for me:
![image](https://github.com/user-attachments/assets/d6b76bec-a629-440c-a826-496f270d76e1)
-->
![image](https://github.com/user-attachments/assets/1c38ea7d-e36b-4fb4-8b2f-1363d7e2eaa1)

could you please provide how it looks for you in DOM? 
if aria-details appears for you as well, it is a question why screenreader doesn't read text from 'simple-text', i guess it should.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-details

EdsonT-Systems:
Hi @tshimber , I just replicated the same example you provided, removing the other properties to focus on the 'aria-detailed-id', and when the screen reader focuses on the input, it reads 'has details blank'."

Result:

Template:
![image](https://github.com/user-attachments/assets/6a66b5dc-2aa7-4597-b31b-94c2c033eeec)

Ui and DOM:
![image](https://github.com/user-attachments/assets/fa490d2b-190c-4b23-9acb-0289f8fe0c9c)


The properties are theoretically be well linked in the DOM, but the NVDA screen reader announces 'has details blank' when the input field is on focus, which didn’t happen with properties like 'invalid' that the screen reader does read.

EdsonT-Systems:
Did you manage to test it with a screen reader? @tshimber

tshimber:
[...]

PR #2352, opened 2024-09-12
aria details vs aria describedby fixed for text field

tshimber:
Fixes https://github.com/telekom/scale/issues/2349
Accessibility fix for text field. In normal case helper text is provided and its id is being set as aria-describedby. Extra details can be provided in aria-details-id, it is being set as aria-details. But for cases when developers dont wanna use helperText for some reason and provide important info with aria-details-id, we set it as aria-describedby, 'cause its more convenient/accessible for final user.


packages/components/src/components/text-field/text-field.tsx
```diff
@@ -275,9 +278,11 @@ export class TextField {
             disabled={this.disabled}
             readonly={this.readonly}
             autocomplete={this.inputAutocomplete}
-            {...ariaDetailedById}
             {...ariaInvalidAttr}
-            {...(this.helperText ? ariaDescribedByAttr : {})}
+            {...(this.helperText || this.ariaDetailedId
+              ? ariaDescribedByAttr
+              : {})}
+            {...(this.helperText && this.ariaDetailedId ? ariaDetailsAttr : {})}
```
amir-ba, line 285: This seems to be a duplicate. One of the ternary operates should be enough

tshimber: @amir-ba I've done by analogy: each {} block for one property. they both can be set up to conditions
image

tshimber: i mean ternary would be good if it was either ariaDescribedBy or ariaDetails up to conditions. but it can be none, one or both of them

amir-ba: I see , then lets keep it as is

amir-ba (approved):

Issue #2360, opened 2024-10-23
Error: Cannot set properties of undefined (setting 'value')


namng050185:
I use scale-dropdown-select but when fill value from variable then error happen. Please see screenshot and help me

![Screenshot 2024-10-23 at 21 02 10](https://github.com/user-attachments/assets/8e4e51e9-e8fe-42fa-b1e7-a8c6001583c2)
![Screenshot 2024-10-23 at 21 02 06](https://github.com/user-attachments/assets/78b6126b-581d-46c2-9bf5-2bd7ee4f6685)

amir-ba:
@namng050185, welcome to the Scale community.  And thanks for reporting your issue. But please provide a code sandbox where your issue is reproducible. And also provide a description of current behavior and the expected behavior. A screenshot alone will not allow us to you in any way.

namng050185:
I think i found issue in dropdown select: i set debug in file "scale-dropdown-select_13.entry.js"  Function appendInputHidden  call in componentDidLoad run after function updateInputHidden run. So error happen by that. And issue only happen when set value from variable in Angular project, Fixed value not happen

PR #2361, opened 2024-10-28
fix: hidden input in dropdown-select fixed

tshimber:
Fixes https://github.com/telekom/scale/issues/2360

Check for hiddenInput el added.


amir-ba (approved):
LGTM

Issue #2363, opened 2024-11-04
Telekom header nav-item: active=false works incorrectly


tshimber:
**Scale Version**
3.0.0-beta.153

**Framework and version**
Vue3

**Current Behavior**
Main menu items remain active having active=false. For proper work active prop should be removed completely, which is problematic with dynamic binding
![image](https://github.com/user-attachments/assets/87e40001-a7bc-4baf-bcf8-d0b08ba3d550)

**Expected Behavior**
Main menu items are not active having active=false.

**Code Reproduction**
```
<scale-telekom-nav-item active="false">
  <a href="#">Topic Three</a>
</scale-telekom-nav-item>
```

**Desktop (please complete the following information):**
Chrome

PR #2364, opened 2024-11-04
active=false in nav-item fixed

tshimber:
Fixes https://github.com/telekom/scale/issues/2363
Css selector logic fixed, so nav item won't look active having active=false


amir-ba (approved):
LGTM!, this helps in keeping the attributes' consistency.

PR #2374, opened 2024-12-04
fix: check for querySelector value added

tshimber:
should fix this error (must happen on comp init, while props are coming befor DOM is ready):
image

in all other places in the component, where querySelector is used, this check for undefined exists, so I've added here as well


amir-ba (approved):
LGTM!, good to see you back here :)

PR #2375, opened 2024-12-12
Fix(2353): ScaleButton disabled propagating

glebbo-dev:
The accessibility issue was fixed by removing disabled attribute from scale-button when disabled status changes to false.

### DIFF
--- a/packages/components/src/components/button/button.tsx
+++ b/packages/components/src/components/button/button.tsx
@@ -201,6 +201,12 @@ export class Button {
       this.disabled && 'disabled'
     );
 
+    if (!this.disabled) {
+      this.hostElement.removeAttribute('disabled');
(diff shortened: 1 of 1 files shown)


packages/components/src/components/button/button.tsx
```diff
@@ -201,6 +201,12 @@ export class Button {
       this.disabled && 'disabled'
     );
 
+    if (!this.disabled) {
```
amir-ba, line 204: IMHO we can achieve the same by using the Stencil [reflect ](https://stenciljs.com/docs/properties#reflect-properties-values-to-attributes-reflect)attribute for the 'disabled' prop

amir-ba:
LGTM, thanks for the change. I will merge this

Issue #2376, opened 2024-12-12
ScaleDataGrid not displayed when there is null value in rows prop


VukDju:
**Scale Version**
3.0.0-beta.56

**Framework and version**
React 18.3.1 with @telekom/scale-components-react

**Current Behavior**
When there is a null value within one of the arrays of the `rows` property, DataGrid is not displayed at all.

**Expected Behavior**
Instead of not displaying the DataGrid at all when there is a null value in our data, the table cell containing the null value should be rendered with some specific styling, so that the user knows we are missing data.

An example of how it should like like when there is a null data can be seen on the image bellow:

![Example of solution for ScaleDataGrid with nulls](https://github.com/user-attachments/assets/05e3cf9d-b790-4c0b-b75d-1a384277fe2e)

**Code Reproduction**
Try it by setting prop `rows` on this demo page to variable `rowsMockupWithNulls` and reload the demo page:

[Scale and React18 with Wrapper Package](https://codesandbox.io/p/sandbox/scale-components-react18-with-wrapper-forked-9fpnx6)

NOMADE55:
I can work on this one. I'll provide a fix in short.

VukDju:
The problem persist if the value in the table cell is `undefined ` also. In addition to that, same error appears for columns with following types: 'date', 'number', 'text'.

NOMADE55:
@VukDju do you have an example of this not working with `date`, `number`, and `text`?

Because I've seen this only breaking with `undefined `and `null`.  
This happens because it tries to transform to string the content's cell.  `undefined` and `null` can't be transformed to string hence the issue.

VukDju:
@NOMADE55  I already gave an example for `text` in the first post for this issue.

Here is the example of component crashing for number:
[Number column null - ScaleDataGrid crashes](https://codesandbox.io/p/sandbox/scale-components-react18-with-wrapper-forked-6pm2hd)
[Date colum null - ScaleDataGrid crashes](https://codesandbox.io/p/sandbox/scale-components-react18-with-wrapper-xqqkrv)

NOMADE55:
@VukDju the issue comes from the `null`/`undefined` value, disregarding it's type. 

A method within the component makes calculation on which cell has the longer content. For this, the values are transformed to string, but this is not possible with `null ` or `undefined` values.

I've pushed a PR to fix this: #2377

VukDju:
[...]

PR #2377, opened 2024-12-19
fix(data-grid): prevents null or undefined break max content length calculation

NOMADE55:
# What does this PR do?
- Fixes issue https://github.com/telekom/scale/issues/2376
- Prevents `null` or `undefined` values from breaking data-grid component.

## Quick summary
Data Grid component calculates the _longest_ row. For this it relies on `.toString()` [method to obtain its length.](https://github.com/telekom/scale/blob/8ce3a450c2bf023dac4c3d5273279f5a7781815a/packages/components/src/components/data-grid/data-grid.tsx#L383C1-L383C64)
Both `null` and `undefined` lack the former one, and will break the component after invoking the method.

## What has been done
- Add to 0 as default-
- optional chaining to prevent these values from breaking the component.

## Some Observations
- This is my first PR here 👋🏻 so if I'm missing something let me know. 
- I'm still figuring lerna and this monorepo out, so any observations are welcome.
- Tested locally in a mock project.

### DIFF
--- a/packages/components/src/components/data-grid/data-grid.tsx
+++ b/packages/components/src/components/data-grid/data-grid.tsx
@@ -380,7 +380,7 @@ export class DataGrid {
     let maxLength = 0;
     let longestContent;
     rows.forEach((row) => {
-      const length = row[columnIndex].toString().length;
+      const length = row[columnIndex]?.toString()?.length || 0;
       if (length > maxLength) {
         longestContent = row[columnIndex];
         maxLength = length;


amir-ba (commented):
@NOMADE55 thanks for this change but IMHO we still need to consider the sorting as well. 
it would be enough for now to add optional chaining (`.?`) to the `sortTabe` function on line 473-474 and 479-480 
```js
  const textA = a[columnIndex]?.toLowerCase();
  const textB = b[columnIndex]?.toLowerCase();
```

NOMADE55:
@amir-ba Sorry, maybe I'm not following, the error seems to happen when trying to get the longest content. 
If a value is undefined or null, shouldn't its length be zero?

amir-ba:
Changes here was ported to [2395](https://github.com/telekom/scale/pull/2395) due to inactivity and will be closed.

Issue #2080, opened 2023-07-31
contrast ratio scale-segmented-button


tk-network:
**Scale Version**
3.0.0-beta.137

**Framework and version**
vue 3.2.45

**Current Behavior**
In our accessibility report, we have problems with the ```scale-segmented-button```. According to this, the contrast ratio is not correct in dark mode.

![image](https://github.com/telekom/scale/assets/93334942/aab859d1-c122-4f29-a64f-7f77c869b615)

**Desktop (please complete the following information):**
 - OS: MAC OS
 - Browser chrome
 - Version 112.0.5615.137

felix-ico:
@JoschaIco this is another UI/UX issue rather than code

JoschaIco:
Hi @tk-network, this indeed seems to be an issue. Please notice that the segmented button is still in beta and has no accessibility approval yet. We are considering your findings while making the last iterations.

tk-network:
@amir-ba, I have just updated to version 3.0.0-beta.154. The issue still exists.

![image](https://github.com/user-attachments/assets/d11357eb-d276-4b53-8918-de0b3bf99e0e)

amir-ba:
should be fixed with [#2385](https://github.com/telekom/scale/pull/2385)

PR #2385, opened 2025-03-10
fix(segmented-button): contrast ratio scale segmented button (#2080)

glebbo-dev:
Fixed accessibility issue in segment buttons (#2080) by changing selected color to background magenta and foreground white. Also changed background color of disabled selected button.


amir-ba (approved):
LGTM, tnx

Issue #2379, opened 2025-01-13
Visual Glitch on Mobile Nav Menu on Firefox browser


AC40:
**Scale Version**
3.0.0-beta.154

**Framework and version**
Plain HTML and Scale Components

**Current Behavior**
The chevrons which indicate the option to drill down or up in the mobile nav menu are out of place by default. They sometimes, inconsistently reset. A way to force them into their place is, weirdly, hovering (not clicking or interacting, just hovering) over them with the element selector. This makes them snap into place. But, whenever the menu is opened again, they are out of place.
A screen recording from the Firefox browser (the only browser I could find, where the issue occurs):
https://github.com/user-attachments/assets/6b6131fb-d36b-436e-9fa2-0cb46e75914f

**Expected Behavior**
As in other browsers, the chevrons should be in place always not not glitch around.
A screen recording from the Chrome browser (works in Edge as well):
https://github.com/user-attachments/assets/e4befd1e-fe99-4f45-b010-828ca7fc0cbb

**Code Reproduction**
Any of the official scale examples with multiple levels work (e.g.: [examples/telekom-header-samples/mega-menu](https://github.com/telekom/scale/tree/dac9e93f2f30ef18d84d586a74b200476646148a/examples/telekom-header-samples/mega-menu)
```HTML
<scale-telekom-header>
    <scale-telekom-nav-list variant="functions" slot="functions" alignment="right">
      <scale-telekom-nav-item class="burger-item">
        <button>
          <scale-badge>
            <scale-icon-action-menu></scale-icon-action-menu>
          </scale-badge>
        </button>
        <scale-telekom-nav-flyout class="mobile-nav-flyout">
          <!-- the mobile navigation structure goes here -->
          <scale-telekom-mobile-flyout-canvas app-name="Standard Version" app-name-link="/foo">

            <scale-telekom-mobile-menu slot="mobile-main-nav" app-name-link="/bar">
                <scale-telekom-mobile-menu-item active>
                  <a href="#">Topic One</a>
                  <scale-telekom-mobile-menu-item active slot="children">
                    <a href="#">Second Level</a>
                    <scale-telekom-mobile-menu-item slot="children">
                      <a href="#">Third Level</a>
                    </scale-telekom-mobile-menu-item>
                  </scale-telekom-mobile-menu-item>
                </scale-telekom-mobile-menu-item>
              </scale-telekom-mobile-menu>
          </scale-telekom-mobile-flyout-canvas>
[...]

PR #2387, opened 2025-04-25
fix(telekom-mobile-menu-item): fixed icon placing

Ibrahim381-dev:
Bugfix: I fixed issue number #2379 

The telekom-mobile-menu-item was placed in the wrong position in the firefox browser.
I fixed the positioning to also work in the Firefox Browser by changing a css position-property.


amir-ba (approved):
LGTM

PR #2391, opened 2025-06-04
feat(dropdown-select): add clear button to improve user experience

Ibrahim381-dev:
I added an optional feature to the dropdown-select component, which allows clearing the selected dropwdown element by klicking an 'x' icon.


amir-ba (commented):
Thanks for the pr but please run a local build and also commit the changes to the readme file that gets generated. In addition please extend the storybook documentation and add the new property

PR #2402, opened 2025-06-30
fix(helper-text): add aria-hidden to helper icons

KrukowskiJ:
This PR improves accessibility by adding aria-hidden="true" to decorative icons in <scale-helper-text>. It prevents screen readers from announcing purely visual icons.

### DIFF
--- a/packages/components/src/components/helper-text/__snapshots__/helper-text.spec.ts.snap
+++ b/packages/components/src/components/helper-text/__snapshots__/helper-text.spec.ts.snap
@@ -7,7 +7,7 @@ exports[`Helper Text should render danger text with danger icon if dange variant
       <span part="text">
         <slot></slot>
       </span>
-      <scale-icon-alert-error></scale-icon-alert-error>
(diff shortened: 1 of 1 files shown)


KrukowskiJ:
Also, I just noticed that setting the decorative prop to true for helper text icons in the code might achieve the same accessibility effect (hiding them from screen readers) and could potentially be better solution, compared to directly adding aria-hidden="true"

KrukowskiJ:
Change done, you can run pipeline

amir-ba:
The unit tests still need to be adapted in order to have a green pipeline. 
Also, I suggest you add the `decorative` attribute as a boolean (present with no value) e.g.

```html
<scale-icon-alert-information decorative />
```

This will be more conform with how we are setting this in other places. This is to only stay consistent throughout the codebase and will work the same as setting the attribute to `true`

amir-ba:
@KrukowskiJ thanks for the quick change. can you also please adjust the following test

src/components/helper-text/helper-text.spec.ts
 you need to add the `docrative` in the expected snapshot

PR #2404, opened 2025-07-06
fix(utils): update event handling for checkbox and radio button

PapeThePope:
Fixed event handling for checkbox and radio button components to use 'checked' attribute instead of 'value', enabling proper v-model functionality in Vue.js applications. The components were previously not compatible with Vues v-model directive because they were emitting events with the wrong attribute.

### DIFF
--- a/packages/components-vue/src/vue-component-lib/utils.ts
+++ b/packages/components-vue/src/vue-component-lib/utils.ts
@@ -8,7 +8,9 @@ export const createCommonRender = (tagName: string, eventNames: string[] = []) =
         ...listeners,
         [eventName]: (event: CustomEvent<any>) => {
           let emittedValue = event.detail;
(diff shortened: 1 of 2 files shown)


amir-ba:
Thank you, @PapeThePope, for taking the time and providing the fix, but I should mention that the `@telekom/scale-component-vue` that you are using is deprecated. Vue.js now has native support for Web Components and you can use the main package and this issue would be solved natively. We are not checking the functionality of the  `@telekom/scale-component-vue` package library explicitly and do not plan to maintain the proxies and wrappers in them. Therefore, if you can use the main package (`@telekom/scale-components`) for your project. I suggest you do that.  [here is the official guide](https://vuejs.org/guide/extras/web-components.html) for using Web components in Vue.

Issue #2400, opened 2025-06-25
Infinite recursion in ScaleSegment react component. onClick handler is called indefinitely.


mwyrembl:
I could debug the issue to this change in version 156 of scale components:

https://github.com/telekom/scale/blob/ba9974e3052e8fb36bb052f66fb33d5e19cf00a7/packages/components/src/components/segment/segment.tsx#L84-L90

I try to make a short reproducable example to show the issue for further analysis.
Happens with React 19.1.0 and scale components 156.
Downgrade to 155 solved the problem.

mwyrembl:
Here is the example: https://codesandbox.io/p/sandbox/quizzical-brook-rjz83l
Please have a look.

mwyrembl:
@amir-ba Unfortunately version 157 did not fix the issue. I updated the example to 157 and when i click a segmented button it still produces an infinite recursion of click events in the console log.

mwyrembl:
@obstmi Because I need to control the selection states of the segments. Then why does this prop even exist if you're not allowed to use it? Emitting a click event on a component re-render seems odd to me.

obstmi:
@mwyrembl  Of course, the solution isn’t ideal. The correct approach would be to control the state of the segments exclusively through the parent component (ScaleSegmentedButton) and manage it via a prop.
However, this wasn’t how the original implementation of the SegmentedButton was done, which made it impossible to directly change the selected state of the segments. The state of the segments was controlled by the segments themselves rather than by the parent component.

To still allow programmatic control of the segments, we implemented an additional event in the child component to avoid a complete refactoring (and thus a breaking change). Unfortunately, this led to an infinite loop.
The challenge for the fix in the current release was now to keep the original functionality (user clicks on a segment) while also enabling programmatic control of the selected state (which is not a prop, not even a forbidden one) — as required in your use case.

In Vue, this works without issues since changing the selected state doesn’t trigger a full re-render:
https://codesandbox.io/p/devbox/laughing-danilo-jg7hdj

For React, I have two ideas: either, as already suggested, skip the additional setting of the selected state (if you only need the segment’s ID, you already have it in the click handler), or use useRef instead of useState, as shown in this example:
https://codesandbox.io/p/sandbox/quizzical-brook-forked-93jkk3

Does that help?

PR #2421, opened 2025-08-12
fix:#2400 infinite recursion in scale segmented button

obstmi:
This fixes [#2400](https://github.com/telekom/scale/issues/2400), which was caused by an infinite loop of emitting and handling the scale-change event.
With this solution, emitting the event now depends on whether the state change was triggered by a user click or by a programmatic change of the selected attribute. In the latter case, emitting the scale-change event is suppressed.


amir-ba (approved):
LGTM tnx

PR #2422, opened 2025-08-15
fix: #2400 infinite recursion in scale segmented button v2 (trial)

obstmi:
This fixes https://github.com/telekom/scale/issues/2400, which was caused by an infinite loop of emitting and handling the scale-change event.
**Important:** This is an alternativ approach to my first solution (#2421 [)](https://github.com/telekom/scale/pull/2421) using the controlled component pattern. It is intended to remedy the fact that in some cases (when the selected state of the segments is changed by the parent), no scale-change events are emitted.
The idea behind is, that all selected-states are controlled and changed by the parent (scale-segmented-button). The child only emits events.

So far it does not work very well: E.g.:
- segment-button.html   -> not ok
- segment-button-2166.html   -> ok
- segment-button-2400.html   -> not ok

Another disadvantage is that selected state changes can now only be made via JavaScript and not by directly modifying the segments. This makes this change a breaking change and, in my opinion, unsuitable as a solution.


obstmi:
Closed, as it was only a trial.

PR #2425, opened 2025-08-27
fix(data-grid): fix overflowing table header text when title space reduced

popadicbranislav:
Fixing the issue with overflowing data-grid table header text when width of the column is reduced.

The sort indicator icons were moved out of the text element and removed their absolute positioning as well as it was never reset and thus the icons were not visible. This I discovered by accident, not sure if previous behaviour was intentional or not, but this seems to work better. 
New position of sort indicator icons is after the header text which also seems to work better as it reduces the move of the text when icon appears.

### DIFF
--- a/packages/components/src/components/data-grid/data-grid.css
+++ b/packages/components/src/components/data-grid/data-grid.css
@@ -169,6 +169,12 @@
   color: var(--telekom-color-text-and-icon-additional);
 }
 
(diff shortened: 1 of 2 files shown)


amir-ba (approved):
Thanks for this, I think the change of the flag position was a good idea as well.

PR #2426, opened 2025-09-01
fix(data-grid): fix issue with table header overlapping the table body

popadicbranislav:
The issue appears when the last on scroll event effect to update the scroll position of the header doesn't happen. It appears as it will not render until the next redraw, sometimes keeping the header stuck over table values when using freeze-headers option and scrolling back to top.

Proposed solution removes the use of on scroll event and utilizes the `position: sticky` for the table header.


amir-ba (approved):
Why we did not do this with Sticky and made the expensive computation for the scroll, makes me question lots of stuff. But thank you for the nice fix. LGTM. I will merge it.

Issue #2413, opened 2025-07-22
scale-telekom-header: lang-switcher does not meet the minimum target size


SystemTeamMagentaBusiness:
**Scale Version**
[3.0.0-beta.156](https://www.npmjs.com/package/@telekom/scale-components/v/3.0.0-beta.156)

**Framework and version**
[Nuxt 4.0.1](https://www.npmjs.com/package/nuxt/v/4.0.1)

**Current Behavior**
Each element in the scale-language-switcher has a size of 13.77x27 px.

**Expected Behavior**
Interactive elements should be easy to activate via pointer input (such as mouse or touch) without users accidentally activating a neighboring element. For users who have difficulties with fine motor skills or tremble, it is difficult to hit small targets accurately. Sufficient size or distance between targets reduces the likelihood of accidentally activating the wrong control element.

Thus, the scale-language-switcher should have a [minimum size of 24x24 px](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum).

**Code Reproduction**
https://telekom.github.io/scale/?path=/docs/components-telekom-brand-header-navigation--mega-menu

The code example links to the official documentation.
It demonstrates the scale-telekom-mega-menu.
Please use the DevTools to inspect the elements.

**Desktop:**
- OS: Windows 11
- Browser: Google Chrome
- Version 138.0.7204.158

PR #2430, opened 2025-10-01
fix-language-switcher: #2413

Ricardo-Tele:
The interactive elements in the scale-language-switcher previously had a size of 13.77 × 27 px, which made them difficult to activate, especially for users with motor impairments.

Fix:

Increased the minimum size of each interactive element to 24 × 24 px.

Ensures compliance with accessibility best practices by reducing the risk of accidental activations.


amir-ba:
will close due to duplicate

PR #2433, opened 2025-10-20
fix: update styles of input components to prevent overflow

popadicbranislav:
Issue: date picker and text field components label overflows the component /wraps to multiple lines in case of very long text provided for the label.

image

image


popadicbranislav:
Not sure why `display: flex` was used for the label, so I removed it.

amir-ba (approved):

PR #2434, opened 2025-10-22
fix: #2406 hide label visually a11y

### LINKED ISSUE #2406: scale-text-field: hide-label-visually=true causes accessibility name is also no longer set
**Scale Version**

[3.0.0-beta.156](https://www.npmjs.com/package/@telekom/scale-components/v/3.0.0-beta.156)

**Framework and version**

[Nuxt 4.0.1](https://www.npmjs.com/package/nuxt/v/4.0.1)

**Current Behavior**
If we set the prop "hide-label-visually=true" the scale-text-field component looses the accessibility name.

**Expected Behavior**
If we set the prop "hide-label-visually=true" the scale-text-field component should still have an accessibility name set.

**Code Reproduction**

https://codesandbox.io/p/devbox/scale-text-field-hide-label-visually-q2tww9

The code example demonstrates two scale-text-field components.
The first one does not show the label (by setting hide-label-visually=true).
The second field shows the label (default behavior).
Please use the DevTools to inspect the elements.

**Desktop:**
 - OS: Windows 11
 - Browser: Google Chrome
 - Version 138.0.7204.158

obstmi:
This PR fixes #2406 .
The issue was caused by the label being hidden using CSS. However, this completely removes the label from the accessibility tree.
As solution, the input field references the corresponding label using “aria-labelledby.” .

### DIFF
--- a/packages/components/src/components/text-field/__snapshots__/text-field.spec.ts.snap
+++ b/packages/components/src/components/text-field/__snapshots__/text-field.spec.ts.snap
@@ -3,10 +3,10 @@
 exports[`TextField should match snapshot 1`] = `
 <scale-text-field label="test">
   <div class="text-field text-field--type-text text-field--variant-informational">
-    <label class="text-field__label" htmlfor="input-text-field-0">
+    <label class="text-field__label" htmlfor="input-text-field-0" id="input-text-field-0-label">
       test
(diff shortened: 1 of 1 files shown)


amir-ba (approved):
LGTM!

Issue #2417, opened 2025-07-22
scale-date-picker: missing reactivity in date-picker-title ("Pick a date")


SystemTeamMagentaBusiness:
**Scale Version**
[3.0.0-beta.156](https://www.npmjs.com/package/@telekom/scale-components/v/3.0.0-beta.156)

**Framework and version**
[Nuxt 4.0.1](https://www.npmjs.com/package/nuxt/v/4.0.1)

**Current Behavior**
The scale-date-picker does not update its title ("buttonLabel") if the localization changes.

**Expected Behavior**
If the localization changes the scale-date-picker should also update its title ("buttonLabel") like the other fields do.

**Code Reproduction**
https://codesandbox.io/p/devbox/vvqc9p

1. Open the link provided above
1. Change the localization via the buttons "EN" or "DE"
1. Open the scale-date-picker and look for the title ("buttonLabel"), month names, day names and everything else is translated

**Desktop:**
- OS: Windows 11
- Browser: Google Chrome
- Version 138.0.7204.158

PR #2438, opened 2025-11-11
Fix/date picker buttonlabel(#2417)

Ricardo-Tele:
Changes Made:

Extracted all manual DOM text updates into a new method updateLocalizedDomBits().

Added @Watch('localization') to re-apply localized text dynamically when the prop changes.

Updated button labels, weekday names, "today" suffix, and custom calendar heading.

### DIFF
--- a/packages/components/src/components/date-picker/date-picker.tsx
+++ b/packages/components/src/components/date-picker/date-picker.tsx
@@ -255,7 +255,7 @@ export class DatePicker {
   @Watch('localization')
   onLocalizationChange() {
     if (this.duetInput && this.localization) {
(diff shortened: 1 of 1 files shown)


packages/components/src/components/date-picker/date-picker.tsx
```diff
@@ -508,4 +477,86 @@ export class DatePicker {
       </Host>
     );
   }
+  private updateLocalizedDomBits = () => {
```
amir-ba: The function name seems to be confusing and not representative of the work it does. 
change it to `updateDomOnLocalizationChange`

PR #2444, opened 2025-11-29
Fix/button disabled false a11y

Ricardo-Tele:
Fixes an issue where disabled=false was still rendered, causing screen readers to treat the button as unavailable


packages/components/src/components/date-picker/date-picker.tsx
amir-ba: please remove the changes notrelevent for this PR . 

just revert this file in this branch

amir-ba (changes requested):

amir-ba:
closed in favor of #2447

PR #2468, opened 2026-02-27
feat: add hide-label-visually prop for scale-date-picker and scale-textarea

### LINKED ISSUE #2441: Add hide-label-visually=true prop for scale-date-picker and scale-textarea
**Describe the feature request**

Add support for a `hide-label-visually="true"` prop to both **`scale-date-picker`** and **`scale-textarea`**.
This prop would visually hide the label while keeping it available to assistive technologies, ensuring the component remains accessible without showing the label in the UI.

**Describe the use case**

Some UI designs require the label to be available for screen readers but not displayed visually—for example, when the label is replaced by a placeholder, icon, or when the layout already conveys enough context.
Other components in the Scale library already support `hide-label-visually`, so adding it to `scale-date-picker` and `scale-textarea` would provide consistency and allow teams to follow accessibility best practices without custom workarounds.

**Describe alternatives you've considered**

* Manually hiding the label using CSS (e.g., utility classes). This is less reliable, may not follow the library’s standards, and risks reducing accessibility if not implemented correctly.
* Wrapping the component with additional custom markup to hide the label. This introduces unnecessary complexity and inconsistencies across codebases.
* Using placeholders instead of labels, which does not fully address accessibility requirements.

**Additional context**

The absence of this prop makes these components behave differently from others in the Scale component set that already support `hide-label-visually`.
Adding this prop would align them with the existing API and improve accessibility and developer experience.

tibisabau:
## Description

Added support for the `hide-label-visually` prop to both `scale-date-picker` and `scale-textarea` components. This prop allows labels to be visually hidden while remaining accessible to assistive technologies, ensuring compliance with accessibility best practices.

The implementation follows the existing pattern used in other Scale components like `scale-text-field` and `scale-dropdown-select`, using `visibility: hidden` on the label element to maintain accessibility while hiding the visual representation.

Closes #2441.

## Checklist if Applicable

- [ ] The tests passed 
- [x] Linting passed 
- [ ] Documentation has been added
- [ ] CHANGELOG.md has been updated

### DIFF
--- a/packages/components/src/components/date-picker/date-picker.css
+++ b/packages/components/src/components/date-picker/date-picker.css
@@ -148,6 +148,10 @@ duet-date-picker .duet-date__toggle:active scale-icon-content-calendar {
   overflow: hidden;
 }
 
+.scale-date-picker--hide-label .date-picker__label {
+  visibility: hidden;
+}
+
 duet-date-picker .duet-date__input::placeholder {
   visibility: hidden;
   color: transparent;
--- a/packages/components/src/components/date-picker/date-picker.css
+++ b/packages/components/src/components/date-picker/date-picker.css
@@ -152,6 +152,10 @@ duet-date-picker .duet-date__toggle:active scale-icon-content-calendar {
   visibility: hidden;
 }
 
+.scale-date-picker--hide-label duet-date-picker .duet-date__input {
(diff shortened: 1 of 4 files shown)


amir-ba:
will close due to it being duplicate of #2471

PR #2481, opened 2026-04-09
fix(css): correct font format fallback

grigoriy-reshetniak:
Due to bug in font order newest font format `woff2` was used as fallback for `woff`. This PR changes the fallback order making `woff` fallback for `woff2`.

`woff2` has a better compression and should be preferred generally. It's also [widely adopted](https://caniuse.com/?search=woff2).

Taking advantage of the opportunity I'd like to ask why `eot` (or even `woff` to certain extent) is needed? IE 11 is retired by MS, marked as dead in [by Browserlist](https://github.com/browserslist/browserslist?tab=readme-ov-file#full-list). It has less than 1% of usage. Is it really worth to have this font format in the repository?

### DIFF
--- a/packages/components/src/global/fonts.css
+++ b/packages/components/src/global/fonts.css
@@ -14,8 +14,8 @@
   font-weight: 900;
   font-style: normal;
   src: url('./fonts/TeleNeoWeb/TeleNeoWeb-Ultra.eot');
-  src: url('./fonts/TeleNeoWeb/TeleNeoWeb-Ultra.woff') format('woff'),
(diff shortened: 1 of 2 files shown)


obstmi:
@grigoriy-reshetniak 
Looks good to me, thank you for your PR. I just run prettier for formatting the code changes. Your suggestion regarding the necessity of the eot files I will discuss with the team.

grigoriy-reshetniak:
@obstmi do you have any updates regarding your decision for `eot` and `woff` files? I could've raise a PR for that.

amir-ba:
Create a PR and we will merge it the legacy browser support point you made is valid

PR #2487, opened 2026-04-22
fix(data-grid/text-cell): pass accessibilityTitle to prefix and suffix icons

### LINKED ISSUE #2194: Accessibility | Data grid Component text cell icon add accessibilityTitle
**Describe the feature request**
When using the data grid component and use the icon prefix/suffix option in the text cell you can set an accessibilityTitle Attribute at the icon html tag or the accessibilityTitle will be set automatically.

**Describe the use case**
When using the data grid component and use the icon prefix/suffix option in the text cell you cannot set an accessibilityTitle Attribute at the icon html tag. Therefor the icon is not accessible for blind users when using a screen reader.

**Describe alternatives you've considered**
I discovered currently a workaround where i add the attribute afterwards.
`
const dataGrid = document.querySelector('#dataGrid')
if (dataGrid) {
  const shadowDataGrid = dataGrid.shadowRoot
  const icons = Array.from(shadowDataGrid.querySelectorAll('scale-icon-communication-available'))
  icons.forEach((icon) => {
    icon.setAttribute('accessibilityTitle', 'verfügbar')
  })
}
`

Ricardo-Tele:
Fixes #2194
The accessibilityTitle  is now forwarded to prefix and suffix icons rendered in data grid text cells. This improves accessibility for screen reader users. Regression tests were added to cover the updated behavior.

### DIFF
--- a/packages/components/src/components/data-grid/cell-handlers/text-cell.tsx
+++ b/packages/components/src/components/data-grid/cell-handlers/text-cell.tsx
@@ -19,6 +19,8 @@ import { Cell } from './cell-interface';
 // editable?: boolean = false
 // iconPrefix?: string eg 'action-download'
 // iconSuffix?: string eg 'action-download'
+// iconPrefixAccessibilityTitle?: string eg 'Download'
+// iconSuffixAccessibilityTitle?: string eg 'Download'
 
 export const TextCell: Cell = {
   defaults: {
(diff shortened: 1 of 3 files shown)


amir-ba (commented):
Thanks, the renderer fix looks right. Could you also update the Data Grid Storybook Text Cell options list to include `iconPrefixAccessibilityTitle` and `iconSuffixAccessibilityTitle` next to `iconPrefix`/`iconSuffix`? Those are new user-facing field config keys and currently aren't discoverable in the docs.

PR #2488, opened 2026-04-27
fix(react-typescript): non event props where missing in ts declaration

christopherGdynia:
### Problem

Since `3.0.0-beta.160`, all component-specific props (`variant`, `label`, `size`, `opened`, etc.) are missing from the TypeScript types of every Scale React component. Event handlers like `onScaleBlur` work correctly, but props like `label` on `<ScaleTextField>` produce TS errors.

**Root Cause:** The `@stencil/react-output-target` dependency was declared as `^1.1.1`, but consumers resolved it to `v1.5.1`. In that version, `StencilReactComponent` changed from 2 to 3 type parameters — the automatic prop extraction via `Partial<Omit<Element, keyof HTMLElement>>` was replaced with an explicit third `Props` generic that defaults to `{}`. Since the generated code only passed 2 type parameters, all component-specific props were lost.

### Solution

- Upgrade `@stencil/react-output-target` to `~1.5.1` (pinned to patch range)
- Regenerate React wrappers — the v1.5.1 generator produces 3-parameter declarations with `JSX.ComponentName` as explicit Props type
- Add a post-build patch to re-export the `JSX` namespace from `dist/components/index.d.ts`
- Add type-level regression tests and CI integration

### Changed Files

| File | Description |
|------|-------------|
| `packages/components/package.json` | Upgraded `@stencil/react-output-target` from `^1.1.1` to `~1.5.1`; added post-build script (`node scripts/patch-custom-elements-types.js`) to the build command |
| `packages/components/framework-targets.ts` | Added required `clientModule: '@telekom/scale-components'` option to `reactOutputTarget()` (mandatory in v1.5.x when `hydrateModule` is set) |
| `packages/components/scripts/patch-custom-elements-types.js` | **New** — Post-build script that appends `export type { JSX } from '../types/components'` to `dist/components/index.d.ts`, bridging the gap between where the v1.5.1 generator imports JSX types and where Stencil exports them |
| `packages/components-react/package.json` | Upgraded `@stencil/react-output-target` from `^1.1.1` to `~1.5.1`; added `test:types` script |
| `packages/components-react/src/components.server.ts` | **Regenerated** — all components now use 3-param `StencilReactComponent<Element, Events, JSX.Component>` instead of 2-param declarations; includes `clientModule` and `properties` mappings for SSR |
| `packages/components-react/__tests__/types.test.tsx` | **New** — Type-level regression test (compile-only, not executed) covering 18 representative components with positive and negative (`@ts-expect-error`) assertions |
| `packages/components-react/tsconfig.test.json` | **New** — TypeScript config for type-checking the test file against the built `dist/` output |
| `.github/workflows/build-pr.yml` | **New job** `type-check-react` — builds Stencil components + React wrappers, then runs `test:types` on every PR |

### Verification

```bash
# Rebuild and verify types
yarn workspace @telekom/scale-components build
yarn workspace @telekom/scale-components-react build
yarn workspace @telekom/scale-components-react test:types  # ✅ passes
```

Before fix:
```tsx
<ScaleTextField label="Name" />  // ❌ TS error: Property 'label' does not exist
<ScaleAlert onScaleBlur={...} /> // ✅ works (events were unaffected)
```

After fix:
```tsx
<ScaleTextField label="Name" />  // ✅ compiles
<ScaleAlert variant="informational" opened headline="Test" /> // ✅ compiles
<ScaleAlert nonExistentProp="x" /> // ❌ correctly errors
```


amir-ba (changes requested):
Thanks for the fix. The React prop typing direction looks right, but I think this needs two changes before merge:

1. The new `type-check-react` job runs `yarn workspace @telekom/scale-components build` without generating assets first. On a clean local run this fails before the React type check with missing `packages/components/src/components/icons/scale-icons.json`. Please add `yarn lerna run generate` before the Stencil build, matching the other build jobs.

2. `packages/components-react/src/components.server.ts` is exported through the `react-server` condition, but the regenerated file starts with `'use client'`. That makes the server entry a client module for RSC-aware bundlers. Please strip that directive from the server wrapper output or otherwise keep the `react-server` export server-safe.

amir-ba:
I ported the pipeline validation steps in to a separate group and also had to reconfigure the post build script to also add the `components` . The previous version was also skipping your type tests , this version will run them.

amir-ba (approved):
LGTM

PR #2490, opened 2026-04-28
fix(tag): keyboard/a11y for tag arrow navigation (#2419)

### LINKED ISSUE #2419: scale-tag: not iterable via arrow keys (screen reader skips the tags)
**Scale Version**
[3.0.0-beta.156](https://www.npmjs.com/package/@telekom/scale-components/v/3.0.0-beta.156)

**Framework and version**
[Nuxt 4.0.1](https://www.npmjs.com/package/nuxt/v/4.0.1)

**Current Behavior**
Navigating to a page containing scale-tag components using the NuxtLink component leads to the behavior, that the scale-tag component is no longer accessible with the arrow keys. Thus, the screen reader skips the content entirely and does not read out the content.

By refreshing the page the scale-tag component is accessible with the arrow keys.

This could be a classic timing race condition between Vue's rendering lifecycle and the asynchronous initialization of web components. This might indicate a potential issue with the scale loader (implementation as a plugin).

**Expected Behavior**
Read-only tags, such as those implemented here, are not in tab order, are not interactive, and do not receive focus, which is correct. However, they should be accessible with the arrow keys for screen reader users.

**Code Reproduction**
https://codesandbox.io/p/devbox/9t9crw

1. Start a screen reader (eg. NVDA)
1. Open the link provided above
1. Click on the link "Goto tags page" in the preview
1. Press the tab key (this should navigate to the Add scale-button)
1. Press the left arrow key (jumps directly to the heading and skipping the tag)
1. Refresh the page, repeat step 2-4 again (now the Screen Reader will read out the tag by spelling each letter backwards)

**Desktop:**
- OS: Windows 11
- Browser: Google Chrome
- Version 138.0.7204.158

amir-ba:
Implement roving tabindex and ARIA role for scale-tag so arrow keys move focus and screen readers announce tags. Add unit and e2e tests.

fixes #2419

### DIFF
--- a/packages/components/src/components/tag/__snapshots__/tag.spec.ts.snap
+++ b/packages/components/src/components/tag/__snapshots__/tag.spec.ts.snap
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Tag should have a link 1`] = `
-<scale-tag href="#">
+<scale-tag aria-selected="false" href="#" role="option" tabindex="0">
   <template shadowrootmode="open">
     <a class="tag tag--color-standard tag--link tag--type-standard" href="#" part="base type-standard color-standard link" target="_self">
       <slot></slot>
@@ -12,7 +12,7 @@ exports[`Tag should have a link 1`] = `
 `;
 
 exports[`Tag should match snapshot 1`] = `
-<scale-tag>
+<scale-tag aria-selected="false" role="option" tabindex="0">
   <template shadowrootmode="open">
     <span class="tag tag--color-standard tag--type-standard" part="base type-standard color-standard">
       <slot></slot>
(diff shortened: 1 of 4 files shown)


amir-ba:
The current implementation does not match #2419's expected behavior and misses real repo markup.

1. `<Host tabindex={0}>` makes every read-only `scale-tag` tabbable, but #2419 explicitly says read-only tags should not be in tab order or receive focus. Disabled tags are also still included in the roving set.

2. `getSiblingTags()` only queries under each tag's immediate parent. In the existing data-grid `TagsCell`, tags are rendered as `<ul><li><scale-tag /></li>...</ul>`, so each tag sees only itself and arrow navigation will not move across the list.

3. `role="option"` / `aria-selected` is emitted on standalone tags without a `listbox` owner, which is invalid ARIA context.

Please add coverage for the actual wrapped `<ul><li><scale-tag>` shape plus disabled/href/dismissable cases, and either move roving behavior to an explicit valid group/container or keep read-only tags out of tab order while fixing the screen-reader issue at the correct ownership boundary. If this PR is intended to close #2419, please also link it with `Closes #2419`.

PR #2492, opened 2026-05-02
fix(dropdown-select): don't steal focus outline when dropdown is closed

### LINKED ISSUE #2428: scale-dropdown-select: no outline on focus
**Scale Version**
@telekom/scale-components@3.0.0-beta.156

**Current Behavior**
When focused, a dropdown select list with initially selected item (<scale-dropdown-select value="x">) only gets a visually perceptible border if the initial value is the first item (or if the "value" attribute is missing). If the initial value is not the first item, the drowdown list gets no outline border. It only appears if I once click the dropdown list (even without selecting an item).

**Expected Behavior**
The dropwdown list always gets an outline border when focused, regardless of the initially selected value/item.

**Code Reproduction**
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
    <title>Dropdown outline</title>
    <link
      rel="stylesheet"
      href="node_modules/@telekom/scale-components/dist/scale-components/scale-components.css"
    />
    <script
      type="module"
      src="node_modules/@telekom/scale-components/dist/scale-components/scale-components.esm.js"
    ></script>
  </head>
  <body>
    <b>first item selected: focus with outline</b><br />
    <scale-dropdown-select label="select" value="caspar" name="test1">
      <scale-dropdown-select-item value="caspar">Caspar</scale-dropdown-select-item>
      <scale-dropdown-select-item value="cedric">Cedric</scale-dropdown-select-item>
      <scale-dropdown-select-item value="cem">Cem</scale-dropdown-select-item>
    </scale-dropdown-select>
    <br /><br />
    <b>second, third, ... item selected: focus without outline</b><br />
    <scale-dropdown-select label="select" value="cedric" name="test2">
      <scale-dropdown-select-item value="caspar">Caspar</scale-dropdown-select-item>
      <scale-dropdown-select-item value="cedric">Cedric</scale-dropdown-select-item>
      <scale-dropdown-select-item value="cem">Cem</scale-dropdown-select-item>
    </scale-dropdown-select>
  </body>
</html>

**Desktop**
 - OS: macOS 15.6.1
 - Browser Firefox
 - Version 142.0.1
 - Browser Safari
 - Version 18.6

**Additional context**
see attached images with CSS styles

Image
Image

er-asm:
## Problem

When `scale-dropdown-select` was initialized with a pre-selected value that was not the first item (e.g. `value="cedric"` with `caspar` being first), the `steal-focus` CSS part was incorrectly applied to the wrapper element on load. This suppressed the focus outline (`outline: var(--focus-outline)`) when the user tabbed to the combobox, breaking keyboard accessibility.

Fixes #2428

## Root cause

`currentIndex` was being set whenever the `value` prop changed (including on initial render via `@Watch('value')`). In `getBasePartMap()`, `currentIndex > -1` adds the `steal-focus` part, which CSS uses to suppress the focus outline. Since `currentIndex` equalled the index of the pre-selected item while the dropdown was closed, the outline was always suppressed.

## Solution

- `currentIndex` is no longer set when `value` changes via `@Watch` while the dropdown is closed
- `setOpen(true)` initializes `currentIndex` to the selected item's index (correct starting point for keyboard navigation)
- `setOpen(false)` resets `currentIndex` to `-1` (removing `steal-focus` when closed)
- `connectedCallback()` intentionally left without setting `currentIndex`

## Changes

- `packages/components/src/components/dropdown-select/dropdown-select.tsx` — fix logic
- `packages/components/src/components/dropdown-select/dropdown-select.spec.ts` — 3 new unit tests covering issue #2428
- `packages/visual-tests/src/dropdown-select.visual.spec.js` — un-skipped visual test suite
- `packages/visual-tests/src/__image_snapshots__/dropdown-select-*` — 16 new visual snapshots (light/dark × standard/disabled/error)
- `packages/visual-tests/setup.js` — fix Express server to bind on `0.0.0.0` so Docker container can reach storybook

## Testing

- Unit tests: all passing (`yarn workspace @telekom/scale-components test --spec`)
- Visual tests: 8 passed, 16 snapshots written (`yarn workspace @telekom/scale-visual-tests jest --forceExit -u`)

cc @maomaoZH @amir-ba


amir-ba (changes requested):
Thanks for the PR @er-asm . Please fix the open-state value sync in `valueChange`. `this.options` is never populated in this component, so `Array.isArray(this.options)` is false and `currentIndex` no longer updates for controlled/programmatic `value` changes while the dropdown is open. That leaves `aria-activedescendant` / the `current` option stale and Enter can re-select the previous option.

Use `readOptions(this.hostElement)` here or otherwise keep `currentIndex` synced, and add a regression test for changing `value` while open.

